PR #49, opened 2015-01-06
PLATFORM-667 directly addressible originals

drsnyder:
It was originally believed that the issue with [favicons](https://wikia-inc.atlassian.net/browse/PLATFORM-667) was the result of the `/revision/latest...` at the end of the favicon link tag. This appears [not to be the case](http://www.google.com/s2/favicons?domain=thelongdark.wikia.com) since [other vignette enabled sites](http://despicableme.wikia.com/wiki/Despicable_Me_Wiki) do not appear to have the same issue.

This PR makes originals directly addressable via `/images/top/middle/original.ext` (drops `/revision/latests`). It also provides some refactoring of `routes` to removed duplicated logic and more tests.

https://wikia-inc.atlassian.net/browse/PLATFORM-667

/cc @nmonterroso

### DIFF
--- a/src/vignette/http/routes.clj
+++ b/src/vignette/http/routes.clj
@@ -18,13 +18,6 @@
             [vignette.util.regex :refer :all]
             [vignette.util.thumbnail :as u]))
 
-(def revisionless-original-route
-  (route-compile "/:wikia:image-type/:top-dir/:middle-dir/:original"
-                 {:wikia wikia-regex
-                  :image-type image-type-regex
-                  :top-dir top-dir-regex
-                  :middle-dir middle-dir-regex
-                  :original original-regex}))
 
 (def original-route
   (route-compile "/:wikia:image-type/:top-dir/:middle-dir/:original/revision/:revision"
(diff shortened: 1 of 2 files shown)


wladekb:
@drsnyder @nmonterroso ping

nmonterroso:
I don't think we need this, since the new route that's defined is the same as the legacy route for originals defined in `vignette.api.legacy.routes`:

```clojure
(def original-route
  (route-compile "/:wikia:path-prefix/:image-type:archive/:top-dir/:middle-dir/:original"
                 {:wikia wikia-regex
                  :path-prefix path-prefix-regex
                  :image-type #"images|avatars"
                  :archive archive-regex
                  :top-dir top-dir-regex
                  :middle-dir middle-dir-regex
                  :original original-regex}))
```

@drsnyder - do you think we still need this? The cleanup in the routes is nice, though, so maybe we can merge that :D

drsnyder:
Ok, this PR has been reduced to the routes cleanup. I took out the revisionless original route in ca1e6f6.

nmonterroso:
Cool. There are some conflicts :(

drsnyder:
merged manually

PR #50, opened 2015-01-12
allow negative offsets in window-crop and window-crop-fixed modes

nmonterroso:
@drsnyder @frankfarmer 
Sometimes, you want to have negative offsets.


drsnyder:
lgtm

PR #52, opened 2015-01-13
external image hotlinking

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-697

Let the app decide how to handle external image hotlinking when requesting originals, instead of relying on string manipulation [here](https://github.com/Wikia/wikia-vcl/blob/master/control-images.vcl#L122). After this goes out, we'll switch the VCL to only send the header.


src/vignette/http/routes.clj
```diff
@@ -137,18 +144,13 @@
         (GET alr/thumbnail-route
              request
              (let [image-params (alr/route->thumb-map (:route-params request))]
-               (if (:unsupported image-params)
-                 (-> (response "unsupported thumbnail route")
-                     (status 307)
-                     (header "Location" (str (env :unsupported-redirect-host "http://images.wikia.com")
-                                             (:uri request))))
-                 (if-let [thumb (u/get-or-generate-thumbnail system image-params)]
```
drsnyder: What was this here for?

nmonterroso: Way back in the day when we didn't support all the legacy url formats, specifically `window-crop` and `window-crop-fixed`. This isn't set anywhere anymore, so no point in keeping it.

drsnyder:
lgtm, :sheep: it

PR #53, opened 2015-01-13
Fix integration setup by separating out the transfer! method into a new protocol

drsnyder:
This PR fixes the integration setup. A while back I added `(i/create-integration-env)` which will create a local integration data directory that could be used with `vignette.storage.local`. I must have broken it back when we moved the `transfer!` logic from a multi-method into a protocol.

This PR moves the `transfer!` method into a separate protocol called `TransferableProtocol ` so that the native `java.io.File` abstraction can be `extend`ed to support the `transfer!` method natively. This is the [expression problem](http://www.ibm.com/developerworks/library/j-clojure-protocols/) which clojure provides a nice solution to. Clojure, I :heart: you.

/cc @nmonterroso @owend just for the love

### DIFF
--- a/src/vignette/storage/local.clj
+++ b/src/vignette/storage/local.clj
@@ -35,13 +35,6 @@
   (list-buckets [this])
   (list-objects [this bucket]))
 
-(extend java.io.File
-  TransferableProtocol
-  {:transfer! (fn [from to]
-                (io/copy from
-                         (io/file to))
-                (file-exists? to))})
-
 (defrecord LocalStoredObject [file stream-close-callbacks]
   StoredObjectProtocol
   (file-stream [this]
@@ -57,8 +50,6 @@
         (close []
           (proxy-super close)
           (doall (map #(% stored-object) (:stream-close-callbacks stored-object)))))))
-
-  TransferableProtocol
   (transfer! [this to]
     (io/copy (file-stream this)
              (io/file to))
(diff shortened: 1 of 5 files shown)


src/vignette/storage/local.clj
```diff
@@ -43,17 +50,19 @@
...
-             (io/file to))
-    (file-exists? to))
   (->response-object [this]
     (let [file (file-stream this)
           stored-object this]
       (proxy [FileInputStream] [file]
         (close []
           (proxy-super close)
-          (doall (map #(% stored-object) (:stream-close-callbacks stored-object))))))))
+          (doall (map #(% stored-object) (:stream-close-callbacks stored-object)))))))
```
nmonterroso: Is this an extra removed `)`?

src/vignette/storage/local.clj
```diff
@@ -35,6 +35,13 @@
   (list-buckets [this])
   (list-objects [this bucket]))
 
+(extend java.io.File
+  TransferableProtocol
+  {:transfer! (fn [from to]
+                (io/copy from
+                         (io/file to))
+                (file-exists? to))})
+
```
drsnyder: This should be `(create-stored-object <file>)` in the integration code.

drsnyder:
@nmonterroso ok, I reduced the fanciness!

nmonterroso:
way less fancy lgtm :ship:

PR #55, opened 2015-01-16
add Content-Disposition header for images

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-632

Add a `Content-Disposition` header so "save as..." dialogs have a hint of the filename to use when saving. Tested in Chrome and Firefox and works as expected. Note that the `filename=` part is kept for user agents that don't support [RFC 5987](https://tools.ietf.org/html/rfc5987), which in my testing it seems that Firefox does not.

### DIFF
--- a/src/vignette/util/image_response.clj
+++ b/src/vignette/util/image_response.clj
@@ -29,7 +29,13 @@
    (error-response code nil)))
 
 (defn create-image-response
-  [image]
-  (-> (response (->response-object image))
-      (header "Content-Type" (content-type image))
-      (header "Content-Length" (content-length image))))
+  ([image image-map]
+    (-> (response (->response-object image))
+        (header "Content-Type" (content-type image))
+        (header "Content-Length" (content-length image))
+        (cond->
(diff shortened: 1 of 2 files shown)


drsnyder:
lgtm

src/vignette/util/image_response.clj
```diff
@@ -29,7 +29,13 @@
...
-      (header "Content-Type" (content-type image))
-      (header "Content-Length" (content-length image))))
+  ([image image-map]
+    (-> (response (->response-object image))
+        (header "Content-Type" (content-type image))
+        (header "Content-Length" (content-length image))
+        (cond->
+          (:original image-map) (header "Content-Disposition"
+                                        (str "inline; filename=\""(:original image-map)"\";"
+                                             "filename*=utf=8' '\""(:original image-map)"\";")))))
```
drsnyder: Want to use `format` here? Might read a little better.

nmonterroso: Would that make the PR too fancy?

(but yeah that's a good idea)

PR #56, opened 2015-01-19
remove filename*= portion of Content-Disposition header

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-751

I think special characters were causing the `filename*=UTF-8' '%s` to break vignette requests in Chrome. Looks like Chrome respects utf-8 characters in `filename="%s"`.


nmonterroso:
@drsnyder - gonna push this out since this is causing problems when rendering images.

drsnyder:
Seems fine to me.

PR #57, opened 2015-01-21
ImageMagick working but returning exit code 1

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-759

Sometimes things work but they tell you they didn't. Underpromise, overdeliver I guess.

### DIFF
--- a/src/vignette/util/thumbnail.clj
+++ b/src/vignette/util/thumbnail.clj
@@ -54,7 +54,9 @@
         args (reduce conj base-command thumb-options)
         sh-out (run-thumbnailer args)]
     (cond
-      (zero? (:exit sh-out)) (io/file temp-file)
+      (or (zero? (:exit sh-out))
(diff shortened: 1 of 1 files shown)


drsnyder:
lgtm. Must be a corner case in imagemagick.

src/vignette/util/thumbnail.clj
```diff
@@ -54,7 +54,9 @@
         args (reduce conj base-command thumb-options)
         sh-out (run-thumbnailer args)]
     (cond
-      (zero? (:exit sh-out)) (io/file temp-file)
+      (or (zero? (:exit sh-out))
+          (and (= 1 (:exit sh-out))
+               (file-exists? temp-file))) (io/file temp-file)
```
drsnyder: Nice to have: a comment about why we are doing this. Not knowing the context, it looks odd with the `1` magic number.

PR #58, opened 2015-01-27
MAIN-3622 Timeline legacy support

drsnyder:
This is another one-off request path supported by the legacy thumbnailer. Here is an [example](http://img2.wikia.nocookie.net/__cb1422289387/television/es/images/timeline/bbe457792492f1b89f21a45aa6ca6088.png) image taken from [here](http://es.television.wikia.com/wiki/Wikivisi%C3%B3n:Administraci%C3%B3n). The paths take the form `/bucket/language/images/timeline/hash.png`. The way we decided to handle this was to allow any of the `:middle-dir` or `:top-dir` paths to be empty.

https://wikia-inc.atlassian.net/browse/MAIN-3622

/cc @nmonterroso @owend (since Nelson might be out)

### DIFF
--- a/src/vignette/api/legacy/routes.clj
+++ b/src/vignette/api/legacy/routes.clj
@@ -41,6 +41,13 @@
                   :middle-dir middle-dir-regex
                   :original original-regex}))
 
+(def timeline-route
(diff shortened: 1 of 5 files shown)


nmonterroso:
LGTM :ship:

PR #59, opened 2015-01-30
only apply density magic when upsizing an image

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-808

The density stuff was originally added for [PLATFORM-608](https://wikia-inc.atlassian.net/browse/PLATFORM-608). 200 JIRA issues later, this is causing issues in SVGs that are downsized from the originals. In those cases, from my testing it seems OK to ignore the density stuff.


drsnyder:
lgtm

PR #60, opened 2015-01-31
MAIN-3701 Add 'Surrogate-Key' header to image requests

drsnyder:
Add `Surrogate-Key` header to all image requests. This will enable more effective purging. The surrogate key is derived from the original image path on ceph including the bucket.

This will need to be integrated into MediaWiki. I'm also considering dialing down the `max-age` that we return so that images aren't cached "forever". That has broader implications though so this should be a good stop gap for handling DMCA requests.

https://wikia-inc.atlassian.net/browse/MAIN-3701

/cc @nmonterroso @ljagiello

### DIFF
--- a/src/vignette/util/image_response.clj
+++ b/src/vignette/util/image_response.clj
@@ -2,11 +2,14 @@
   (:require [clojure.java.io :refer [file]]
             [compojure.route :refer [not-found]]
             [ring.util.response :refer [response status header]]
+            [vignette.media-types :refer :all]
             [vignette.storage.local :refer [create-stored-object]]
             [vignette.storage.protocols :refer :all]
             [vignette.util.thumbnail :refer :all]))
 
-(declare create-image-response)
+(declare create-image-response
+         add-surrogate-header
+         surrogate-key)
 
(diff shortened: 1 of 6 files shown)


src/vignette/util/image_response.clj
```diff
@@ -30,12 +33,30 @@
...
-                                        (format "inline; filename=\"%s\""
-                                                (:original image-map))))))
+   (-> (response (->response-object image))
+       (header "Content-Type" (content-type image))
+       (header "Content-Length" (content-length image))
+       (header "X-Thumbnailer" "Vignette")
+       (cond->
+         (original image-map) (header "Content-Disposition"
+                                       (format "inline; filename=\"%s\""
+                                               (:original image-map)))
```
nmonterroso: shouldn't this also be `(original image-map)`?

nmonterroso:
LGTM

src/vignette/util/image_response.clj
drsnyder: Fixed here 146c33c.

nmonterroso:
LGTM :shipit: :ship: :rocket:

PR #61, opened 2015-02-02
don't let me get in my zone

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-824

add support for `zone` in both legacy and vignette routes. This is the Vignette side of https://github.com/Wikia/app/pull/6135

### DIFF
--- a/src/vignette/media_types.clj
+++ b/src/vignette/media_types.clj
@@ -15,9 +15,9 @@
 
 (defmethod image-type->path-prefix "images" [object-map]
   (let [path-prefix (query-opt object-map :path-prefix)
-        lang (query-opt object-map :lang)
+        zone (query-opt object-map :zone)
         prefix (:image-type object-map)]
-    (clojure.string/join "/" (filter not-empty [path-prefix lang prefix]))))
+    (clojure.string/join "/" (filter not-empty [path-prefix prefix zone]))))
 
 (defmethod image-type->path-prefix :default [object-map]
   (throw+ {:type :convert-error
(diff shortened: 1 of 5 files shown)


src/vignette/media_types.clj
```diff
@@ -15,10 +15,15 @@
   (image-type object-map))
 
 (defmethod image-type->path-prefix "images" [object-map]
+  "
+  path-prefix comes before the image-type, and can be a language code (ex: 'es') or a path (ex: /some/dir)
+  zone is further distinction of what type of image. For instance, 'temp' indicates the image is a temporary image.
+    this is needed because mediawiki supports these in filerepo->getZoneUrl
+  "
```
drsnyder: An example path would be helpful here.

test/vignette/media_types_test.clj
```diff
@@ -25,12 +25,11 @@
...
 
-(def lang-map (assoc latest-map :options {:lang "es"}))
+(def lang-map (assoc latest-map :options {:path-prefix "es"}))
 
-(def prefix-path-map (assoc latest-map :options {:lang "de"
-                                                 :path-prefix "pokemanshop/zh"}))
+(def prefix-path-map (assoc latest-map :options {:path-prefix "pokemanshop/zh/de"}))
 
-(def lang-original-map (assoc original-map :options {:lang "es"}))
+(def lang-original-map (assoc original-map :options {:path-prefix "es"}))
```
drsnyder: Want to add a test for `original-path` that includes the zone to confirm that the correct path on disk is generated?

nmonterroso: Good call

PR #62, opened 2015-02-02
Use the hash of the original path as the surrogate key

drsnyder:
Use the hash of the original path as the surrogate key instead of just the original path. There may be some limitations on the size of the key we can use and this should ensure that we comply.

/cc @pchojnacki @nmonterroso


ljagiello:
Fatal error in travis

drsnyder:
Updated the tests in 7956fbb.

nmonterroso:
LGTM

PR #63, opened 2015-02-05
Log the bad request paths

drsnyder:
This increments a counter and logs the path whenever we drop into the 404/not-found handler in the routes. This should log any path that does not match the legacy or new style URLs.

/cc @frankfarmer @ljagiello @nmonterroso

### DIFF
--- a/src/vignette/http/middleware.clj
+++ b/src/vignette/http/middleware.clj
@@ -1,5 +1,6 @@
 (ns vignette.http.middleware
   (:require [environ.core :refer [env]]
+            [compojure.response :refer [render]]
             [ring.util.response :refer [response status charset header]]
             [slingshot.slingshot :refer [try+ throw+]]
             [vignette.util.image-response :refer :all]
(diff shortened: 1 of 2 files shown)


src/vignette/http/middleware.clj
```diff
@@ -46,7 +47,16 @@
...
   (fn [request]
     (perf/publish {:request-count 1})
     (perf/timing :request-time (handler request))))
+
+(defn bad-request-path
+  []
+  (fn [request]
+    (perf/publish {:bad-request-path-count 1})
+    (log/warn (str "bad-request-path: " (:uri request)))
+    (-> (render "Unrecognized request path!\nSee https://github.com/Wikia/vignette for documentation.\n" request)
```
nmonterroso: any reason we couldn't call `not-found` here?

nmonterroso:
Other than minor comment LGTM :shirt:

src/vignette/http/middleware.clj
drsnyder: `not-found` is generated using a [macro that adds a wrapper](https://github.com/weavejester/compojure/blob/master/src/compojure/route.clj#L44) that seems to interfere with adding this as middleware. I tried that first but had trouble getting it to work, so I just reproduced the core of the function here.

nmonterroso: Ah, ok. I think that happened to me once. LGTM

PR #64, opened 2015-02-05
MWEB-1100 Add legacy routes for interactive maps

drsnyder:
Interactive maps stores images in the following forms which do not map to any of the currently known legacy image paths:

```
/intmap_tile_set_4823/20150205173220!phpZDfa00.jpg
/intmap_tile_set_4823/thumb/20150205173220%21phpZDfa00.jpg/1110x300x5-20150205173220%21phpZDfa00.jpg
/intmap_tile_set_4692/3/1/2.png
```

To support this we need to add explicit legacy support for these paths while also making the path generation code more tolerant of missing fields. 

This PR:

* Adds legacy routes to support interactive maps
* Makes the path generation more tolerant of empty fields
* Adds tests off the new routes

Issue [MWEB-1100](https://wikia-inc.atlassian.net/browse/MWEB-1100) is a side effect of moving all of the legacy traffic to vignette. See https://github.com/Wikia/chef-repo/pull/4197.

/cc @nmonterroso @ljagiello @rogatty

### DIFF
--- a/src/vignette/api/legacy/routes.clj
+++ b/src/vignette/api/legacy/routes.clj
@@ -17,6 +17,7 @@
 (def thumbname-regex #".*?")
 (def video-params-regex #"(?i)v,\d{6},|v%2c\d{6}%2c|")
 (def zone-regex #"\/[a-z]+|") ; includes "archive"
+(def interactive-maps-regex #"intmap_tile_set_\d+")
 
 (def thumbnail-route
   (route-compile "/:wikia:path-prefix/:image-type/thumb:zone/:top-dir/:middle-dir/:original/:videoparams:dimension:offset:thumbname"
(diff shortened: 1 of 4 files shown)


src/vignette/api/legacy/routes.clj
```diff
@@ -49,6 +50,20 @@
                   :image-type #"images"
                   :original original-regex}))
 
+(def interactive-maps-route
+  (route-compile "/:wikia:path-prefix/:original"
+                 {:wikia interactive-maps-regex
+                  :path-prefix #"/\d/\d|"
```
rogatty: The first number currently can be [in range 0-8](https://github.com/Wikia/config/blob/dev/InteractiveMapsConfig.json#L144) but the second one might be higher. It depends on the image size for given zoom. For example /5/23/9.png is a valid path.

src/vignette/api/legacy/routes.clj
```diff
@@ -49,6 +50,20 @@
...
+                  :original original-regex}))
+
+(def interactive-maps-thumbnail-route
+  (route-compile "/:wikia/thumb/:original/:dimension:offset:thumbname"
```
nmonterroso: Does this need `:path-prefix` as well?

drsnyder: Not that I'm aware of. @rogatty do you know if there can be directories between `thumb` and `20150205173220%21phpZDfa00.jpg` in this image:

`/intmap_tile_set_4823/thumb/20150205173220%21phpZDfa00.jpg/1110x300x5-20150205173220%21phpZDfa00.jpg`

If so then we'll need to add the `:path-prefix`.

nmonterroso: We talked about this offline, looks like we don't need this.

src/vignette/api/legacy/routes.clj
drsnyder: Thanks. I change both to `\d+` in 3b5f7c4.

nmonterroso:
LGTM

PR #66, opened 2015-02-06
adding find-images.js script to easily find vignette images on web pages

nmonterroso:
@drsnyder 
Should help with all those "the path bugs have been fixed but we're waiting for Fastly cache to expire" issues we've been having lately.


drsnyder:
Cool, lgtm.

PR #68, opened 2015-02-11
have a blacklist of content-types that we can't thumbnail

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-883

we can't thumbnail video/audio files :(

### DIFF
--- a/src/vignette/http/middleware.clj
+++ b/src/vignette/http/middleware.clj
@@ -21,12 +21,14 @@
               thumb-map (:thumb-map e) ; if present, we'll try to thumbnail the error response
               response-code (or (:response-code e) 500)
               context (assoc (dissoc e :type :thumb-map :response-code) :host hostname)]
-          (perf/publish {:convert-error 1})
-          (println message ":" (:uri request))
-          (log/warn message
-                    (merge {:path (:uri request)
-                            :query (:query-string request)}
-                           context))
+          (when (>= response-code 500)
+            (perf/publish {:convert-error 1})
+            (println message ":" (:uri request))
+            (log/warn message
+                      (merge {:path (:uri request)
+                              :query (:query-string request)
+                              :response-code response-code}
+                             context)))
           (error-response response-code thumb-map)))
       (catch Exception e
         (println (.getMessage e) ":" (:uri request))
--- a/src/vignette/util/thumbnail.clj
+++ b/src/vignette/util/thumbnail.clj
@@ -32,7 +32,7 @@
                   :window-width "window-width"
                   :window-height "window-height"})
 
-(def mime-blacklist #{"audio/ogg"
+(def unsupported-mime-types #{"audio/ogg"
                       "video/ogg"})
 
 (defn route-map->thumb-args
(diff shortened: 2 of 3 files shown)


src/vignette/http/middleware.clj
```diff
@@ -21,12 +21,14 @@
               thumb-map (:thumb-map e) ; if present, we'll try to thumbnail the error response
               response-code (or (:response-code e) 500)
               context (assoc (dissoc e :type :thumb-map :response-code) :host hostname)]
-          (perf/publish {:convert-error 1})
-          (println message ":" (:uri request))
-          (log/warn message
-                    (merge {:path (:uri request)
-                            :query (:query-string request)}
-                           context))
+          (when (>= response-code 500)
```
drsnyder: Hmm. This looked odd at first. Will we ever get something other than a `5xx` here?

drsnyder: Ohh. Nvm. I see why below.

src/vignette/util/thumbnail.clj
```diff
@@ -29,6 +32,9 @@
                   :window-width "window-width"
                   :window-height "window-height"})
 
+(def mime-blacklist #{"audio/ogg"
```
drsnyder: Instead of `mime-blacklist` should this be `unsupported-mime-types` or something similar that indicates more clearly that we aren't supporting thumbnail operations on these types?

src/vignette/util/thumbnail.clj
```diff
@@ -42,7 +48,23 @@
...
+
+(defmethod original->thumbnail LocalStoredObject [resource thumb-map]
+  (let [content-type (content-type resource)]
+    (when (contains? mime-blacklist content-type)
+      (throw+ {:type :convert-error
+               :response-code 404
+               :content-type content-type
+               :thumb-map thumb-map}
+              "unsupported content-type"))
+    (file->thumbnail (file-stream resource) thumb-map)))
```
drsnyder: Why is the logic different here depending on if it's a `File` or a `LocalStorageObject`? I would have thought that may this would be a `:pre` condition or a check of some form.

ljagiello:
@nmonterroso sure you can:
```
ffmpeg -ss 4 -i Kennen_RVideo.ogv -vframes 1 -s 420x270 -filter:v 'yadif' output.png
```

drsnyder:
@ljagiello yes, but we are reluctant to start supporting both ffmpeg and imagemagic unless there is a compelling reason to do so. As far as we know, the old thumbnail also did not support this.

src/vignette/util/thumbnail.clj
nmonterroso: Yeah, that's a good point

src/vignette/util/thumbnail.clj
nmonterroso: Hm, I wanted to use `(content-type)`, but I guess that's not necessary.

nmonterroso: `:pre` is magic!

drsnyder:
lgtm. :sheep: it

PR #69, opened 2015-02-11
add custom format to content disposition header

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-881

so people who save images don't think we didn't apply their custom format.

### DIFF
--- a/src/vignette/util/image_response.clj
+++ b/src/vignette/util/image_response.clj
@@ -4,6 +4,7 @@
             [ring.util.response :refer [response status header]]
             [digest :as digest]
             [vignette.media-types :refer :all]
+            [vignette.util.query-options :refer :all]
             [vignette.storage.local :refer [create-stored-object]]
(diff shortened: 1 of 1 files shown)


drsnyder:
lgtm

src/vignette/util/image_response.clj
```diff
@@ -49,9 +50,10 @@
 
 (defn add-content-disposition-header
   [response-map image-map]
-  (header response-map "Content-Disposition"
-          (format "inline; filename=\"%s\""
-                  (original image-map))))
+  (let [requested-format (query-opt image-map :format)
+        filename (cond-> (original image-map)
+                         requested-format (str "." requested-format))]
```
drsnyder: Will this produce `file.png.webp`?

nmonterroso: Yeah. Seems easier than trying to strip out the old extension, especially since it might not be there.

PR #70, opened 2015-02-17
Switch the HTTP server to Jetty 9

drsnyder:
It's unfortunate, but perhaps providential that http-kit doesn't [arbitrary HTTP methods](https://github.com/http-kit/http-kit/blob/b06ed00dcf058dd1c4d68c61c7cea12c15282cab/src/java/org/httpkit/HttpMethod.java). Jetty does though and since dropwizard also uses Jetty there is a little more incentive to switch to jetty than attempt to change http-kit. We weren't really leveraging any of the potential benefits of http-kit anyway.

We need to support arbitrary HTTP methods to support PURGE. See [PLATFORM-888](https://wikia-inc.atlassian.net/browse/PLATFORM-888).

The current [`ring-jetty-adapter`](https://github.com/ring-clojure/ring/tree/master/ring-jetty-adapter) intends to maintain backwards compatibility with earlier versions of Java that are not compatible with later versions of jetty. Since dropwizard is on the Jetty 9 version I also targeted that version. Someone already created an updated adapter [here](https://github.com/sunng87/ring-jetty9-adapter) which has reasonable activity so I went with it. The changes to the adapter from the original appear to be mostly forward compatibility and performance related.

I also added an optional NCSA access logging capability. This was something that I have missed both in the current vignette and other jetty based clojure projects.

We'll need to be careful in rolling this out. We should bake it in dev with some manual QA (or beef up the integration tests) a little more than normal before we roll this out.

/cc @nmonterroso @pchojnacki @macbre 

TODO:
 * [ ] review and revise the integration tests as necessary


drsnyder:
Got a verbal lgtm from @nmonterroso. I'm also burning this in on vignette-s1 for a few hours. So far so good.

PR #71, opened 2015-02-19
PLATFORM-888 Add support for purging varnish via the PURGE method

drsnyder:
Add support for purging varnish via the PURGE method. To add support a request macro including both `GET` and `PURGE` was added to keep the routing logic to a minimum. As part of this process I further simplified the vignette routes by moving the `:thumbnail-mode` logic into the routes and out of the handler. This exposed the dangling `{:height :auto}` which was also moved out of the handler logic and into `bin/thumbnail`.

To handle the purging a `cache` was added to the `SystemAPI`. A new protocol for purging was also added to provide an interface to this behavior. See `CachePurgeAPI`. It supports `(purge [uri surrogate-key])`. Finally, a fastly implementation for the protocol is provided that allows the credentials to be injected via the environment.

The corresponding changes on the MW side still need to be implemented.

/cc @nmonterroso @macbre @harnash

### DIFF
--- a/bin/thumbnail
+++ b/bin/thumbnail
@@ -80,6 +80,25 @@ cat <<EOF
 EOF
 }
 
+# Test for array containment.
+# Example: $(contains "${A[@]}", "foo") == "y"
+# Returns "y"" when the object is contained in the array and "n" otherwise.
+# http://stackoverflow.com/questions/3685970/check-if-an-array-contains-a-value
+function contains() {
+	local n=$#
+	local value=${!n}
+	for ((i=1;i < $#;i++)) {
+		if [ "${!i}" == "${value}" ]; then
+			echo "y"
+			return 0
+		fi
+	}
+
+	echo "n"
+	return 1
+}
+
+
 OS=$(uname)
 SHORTOPTS="h"
 LONGOPTS="in:,out:,width:,height:,fuzz:,help,fill:,still,mode:,animated,x-offset:,window-width:,y-offset:,window-height:"
@@ -185,10 +204,13 @@ if [ "$OUT"X = "X" ]; then
 	exit 1
 fi
 
+MODES_REQUIRING_HEIGHT=("fixed-aspect-ratio" "fixed-aspect-ratio-down" "thumbnail" "thumbnail-down" "top-crop" "top-crop-down" "window-crop-fixed" "zoom-crop" "zoom-crop-down")
 if [ "$HEIGHT"X = "X" ]; then
-	echo "--height required"
-	usage
-	exit 1
+	if [ $(contains "${MODES_REQUIRING_HEIGHT[@]}" "${MODE}") == "y" ]; then
+		echo "--height required for mode ${MODE}"
+		usage
+		exit 1
+	fi
 fi
 
 if [ "$WIDTH"X = "X" ]; then
--- a/src/vignette/caching/edge/fastly.clj
+++ b/src/vignette/caching/edge/fastly.clj
@@ -1,0 +1,47 @@
+(ns vignette.caching.edge.fastly
+  (:require [clj-http.client :as client]
+            [vignette.protocols :refer :all]
+            [wikia.common.logger :as log]
+            [wikia.common.perfmonitoring.core :as perf])
+  (:use [environ.core]))
+
+(def default-fastly-api-url "https://api.fastly.com")
+
+(def fastly-creds {:id       (env :fastly-api-id)
+                   :auth-key (env :fastly-api-auth-key)
+                   :api-url  (env :fastly-api-url default-fastly-api-url)})
+
+(declare purge-request-url
+         api-params)
+
+(defrecord FastlyAPI [id auth-key fastly-api-url]
+  CachePurgeAPI
+  (purge [this uri surrogate-key]
+    (perf/publish {:purge-count 1})
+    (let [response (client/post (purge-request-url (:fastly-api-url this)
+                                                   (:id this)
+                                                   surrogate-key)
+                                (api-params (:auth-key this)))]
+      (if (not= (:status response) 200)
+        (do (perf/publish {:failed-purge-count 1})
+            (log/warn (:body response) {:path uri :key surrogate-key})
+            false)
+        (do
+          (log/info (format "purged: %s using key %s" uri surrogate-key) {:path uri :key surrogate-key})
+          true)))))
+
+(defn create-fastly-api
+  [creds]
+  (->FastlyAPI (:id creds) (:auth-key creds) (:api-url creds)))
+
+(defn purge-request-url
+  [fastly-api id surrogate-key]
+  (format "%s/service/%s/purge/%s"
+          fastly-api
+          id
+          surrogate-key))
+
+(defn api-params
+  [auth-key]
+  {:headers {"Fastly-Key" auth-key
+             "Accept" "application/json"}})
--- a/src/vignette/caching/edge/fastly.clj
+++ b/src/vignette/caching/edge/fastly.clj
@@ -11,6 +11,10 @@
                    :auth-key (env :fastly-api-auth-key)
                    :api-url  (env :fastly-api-url default-fastly-api-url)})
 
+(defn empty-fastly-creds?
+  [creds]
+  (some nil? (vals creds)))
+
 (declare purge-request-url
          api-params)
 
@@ -24,7 +28,7 @@
                                 (api-params (:auth-key this)))]
       (if (not= (:status response) 200)
         (do (perf/publish {:failed-purge-count 1})
-            (log/warn (:body response) {:path uri :key surrogate-key})
+            (log/warn "vignette fastly purge failed" {:path uri :key surrogate-key :error (:body response)})
             false)
         (do
           (log/info (format "purged: %s using key %s" uri surrogate-key) {:path uri :key surrogate-key})
--- a/src/vignette/core.clj
+++ b/src/vignette/core.clj
@@ -5,7 +5,7 @@
             [vignette.storage.local :refer [create-local-storage-system]]
             [vignette.storage.protocols :refer :all]
             [vignette.storage.s3 :refer [create-s3-storage-system storage-creds]]
-            [vignette.caching.edge.fastly :refer [create-fastly-api fastly-creds]]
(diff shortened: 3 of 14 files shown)


ljagiello:
I wasn't part of this discussion but why you're not using purger (and purger workflow) here ?

bin/thumbnail
```diff
@@ -80,6 +80,25 @@ cat <<EOF
...
+# Test for array containment.
+# Example: $(contains "${A[@]}", "foo") == "y"
+# Returns "y"" when the object is contained in the array and "n" otherwise.
+# http://stackoverflow.com/questions/3685970/check-if-an-array-contains-a-value
+function contains() {
+	local n=$#
+	local value=${!n}
+	for ((i=1;i < $#;i++)) {
+		if [ "${!i}" == "${value}" ]; then
+			echo "y"
```
nmonterroso: is `echo` the same as `return`? Will these add to the actual output of running this script?

src/vignette/caching/edge/fastly.clj
```diff
@@ -0,0 +1,47 @@
...
+  CachePurgeAPI
+  (purge [this uri surrogate-key]
+    (perf/publish {:purge-count 1})
+    (let [response (client/post (purge-request-url (:fastly-api-url this)
+                                                   (:id this)
+                                                   surrogate-key)
+                                (api-params (:auth-key this)))]
+      (if (not= (:status response) 200)
+        (do (perf/publish {:failed-purge-count 1})
+            (log/warn (:body response) {:path uri :key surrogate-key})
```
nmonterroso: should this be a literal string that we can search for, and maybe put `:body` in the context?

src/vignette/core.clj
```diff
@@ -20,6 +21,10 @@
                  :parse-fn #(Integer/parseInt %)
                  :validate  [#(< 0 % 0x10000) "Must be a number between 0 and 65536"]]])
 
+(defn empty-fastly-creds?
```
nmonterroso: should this live in the fastly namespace?

src/vignette/http/routes.clj
```diff
@@ -189,12 +198,28 @@
     (create-image-response file image-params)
     (error-response 404 image-params)))
 
+(declare background-purge)
```
nmonterroso: can we keep `declare`s at the top? makes them easier to find

src/vignette/util/thumbnail.clj
```diff
@@ -74,6 +74,7 @@
                (file-exists? temp-file))) (io/file temp-file)
       :else (throw+ {:type :convert-error
                      :error-code (:exit sh-out)
+                     :out-string (:out sh-out)
```
nmonterroso: I think we used to have this, but I removed it because there is a limit to the size of the message the java library will send using the syslog protocol. When it goes over, the json is partially missing, so it can't be json decoded, and the message is effectively lost (though the raw json string does make it to ES)

bin/thumbnail
drsnyder: No, with `echo` you can capture the value in `$()`. If you `return` or `exit` then you have to check `$?`. Because `$()` is used below the output is captured.

src/vignette/caching/edge/fastly.clj
drsnyder: Oh, I get it. Yeah. Not sure what I was thinking.

src/vignette/core.clj
drsnyder: Fixed in 1249446.

src/vignette/util/thumbnail.clj
drsnyder: Bummer. Ok, I removed it from the logging payload in c59cfbe.

drsnyder:
> I wasn't part of this discussion but why you're not using purger (and purger workflow) here ?

@ljagiello the reason is that the authoritative source of the surrogate key is vignette. I don't want to duplicate the logic for generating the surrogate key in MW. That leaves us 2 options for handling the surrogate key purge: 1) do a `HEAD` on the resource in vignette, get the surrogate key and then enqueue to the traditional purger. Or, 2) send the `PURGE` to vignette and let it handle the implementation details.

1) is not currently viable because the current purger doesn't support surrogate keys through MW. We could add that but there is also a new purger in the works that does support surrogate keys but AFAIK it's not in production. It's also seems like a poor separation of concerns to do it this way. Why should MW be concerned with purging images?

2) is currently the more straight forward option both in terms of the implementation and the architecture. We will send the `PURGE` to vignette and let it handle the details of how that works. For now, we do the purge ourselves but it would be easy to change this to use the platform purger that supports surrogate keys when it's ready.

Hope that helps. Let me know if you see any issue.

ljagiello:
@drsnyder 

1) https://github.com/Wikia/celery-workers/blob/master/celery_workers/purger.py#L60 <- you're talking about that ?

"New purger" is prod ready we're waiting for platform team to finish their part (https://wikia-inc.atlassian.net/browse/PLATFORM-704). @owend is also working on this since we need new purger for internal cache (project should be finished Q1). Creating a new logic and not finish what's currently waiting sounds wrong for me specially if we're going to do it anyway.

2) Sure that make sense I had impression that you want to handle entire purge over vignette but if it will create correct even for purger thats fine.

drsnyder:
> 1) https://github.com/Wikia/celery-workers/blob/master/celery_workers/purger.py#L60 <- you're talking about that ?

Sounds like it.

> "New purger" is prod ready we're waiting for platform team to finish their part (https://wikia-inc.atlassian.net/browse/PLATFORM-704). @owend is also working on this since we need new purger for internal cache (project should be finished Q1). Creating a new logic and not finish what's currently waiting sounds wrong for me specially if we're going to do it anyway.

Most of what I'm implementing will stay in place even if we go that route. I added the direct-to-fastly support (falstly.clj) because the parent ticket of (PLATFORM-888) is currently a P3 that is about to expire.

The fastly part is about 20 lines of Clojure that could be considered a place holder that's easy to swap out. If we want to use the "standard" when it's ready I'm happy to port it over. This PR fills in the rest of the nuts-and-bolts to get us there.

> 2) Sure that make sense I had impression that you want to handle entire purge process over vignette but if it will create correct event for purger thats fine.

That's mostly what this PR is about. I added the `CachePurgeAPI` and the `PURGE` HTTP method support so that when we are ready to use the celery task the core logic won't change only the underlying implementation for purging.

owend:
Couple of questions: Who is working on the PHP part of this?  Is the PHP going to send a request directly to vignette, or is it going to go through the existing perl purger?  

The new purger currently doesn't support multiple end-points, just Fastly.  Once we add that I think Vignette just becomes another end-point, but if Vignette is purging Fastly via surrogate keys, we need to make sure the new/old purger isn't also handling that url, right?

owend:
Also, I'm fine with adding this support to the "old" system.  I think it would allow us to finish the new purger with less pressure due to it being a blocker.

ljagiello:
@owend I believe vignette should provide only API and leave purging to the purger.

Theoretical situation:
- app X want to purge some image + thumbs (sending event to rabbitmq)
- Purger (reading queue) should ask Vignette for surrogate keys and create list of objects to purge
- Purger should send 'PURGE' to all end-points for this event.

If Vignette start doing 'PURGE' we will need to keep 2 places with list of end-points and logic supporting it. We should be open for more end-points not only Fastly.

drsnyder:
> Couple of questions: Who is working on the PHP part of this? Is the PHP going to send a request directly to vignette, or is it going to go through the existing perl purger?

@owend I'm working on the PHP side. PHP will send a `PURGE` request directly to vignette. See the rational as to why above.

> The new purger currently doesn't support multiple end-points, just Fastly. Once we add that I think Vignette just becomes another end-point, but if Vignette is purging Fastly via surrogate keys, we need to make sure the new/old purger isn't also handling that url, right?

I never imagined that scenario. This is what the flow of request looks like:

1. MW needs to purge an image. It will send a `PURGE` request to vignette.
2. Vignette receives the `PURGE` and will either
  1. Call the fastly API directly to purge based on the surrogate key. This is what we are doing now but it's not intended to be a long term solution.
  2. Enqueue the purge request to the standard purger to let it handle the purge to fastly and whatever else needs to be purged.

I think the distinction that needs to be made is that the interface to purging will be in vignette. The implementation details will be abstracted behind vignette. It can either call fastly directly (what this PR provides) or enqueue to the standard purger.

Think of this also as taking more logic out of the monolith. MW will no longer need to know about the layout of files in CEPH purge. It will only need to know that vignette handles purging and there is a method I can use to tell it to purge.

ljagiello:
just $0.02 from me:

> Think of this also as taking more logic out of the monolith. MW will no longer need to know about the layout of files in CEPH purge. It will only need to know that vignette handles purging and there is a method I can use to tell it to purge.

I think a proper approach is to send "pure" PURGE event so MW don't need to know about any other services than purger (purger event queue). It's a purger task to find correct place to purge and construct proper message.

drsnyder:
> I think a proper approach is to send "pure" PURGE event so MW don't need to know about any other services than purger (purger event queue). 

I agree, but we are not yet ready to do that. When we are, I think it would be fine to have a purger request the surrogate key from vignette so that it can do the proper purging.

To be honest, I think it's a lot more efficient (read faster for us) to add this logic in vignette now since the global purger is not ready. It's just a short term solution though.

> It's a purger task to find correct place to purge and construct proper message. 

I agree. And when the purger is ready in production we'll make the necessary changes.

> Vignette should create thumbs and/or provides info about those thumbs. Additional logic like purging it's a different service IMHO.

Again, I agree. But we do not have a purger that currently supports surrogate keys running in production and we have a P3 that is about to expire.

The migration to the new [purger](https://wikia-inc.atlassian.net/browse/PLATFORM-704) has not started development. This PR also lays the ground work for making it easy to get the surrogate key from vignette (e.g. from a `HEAD`).

drsnyder:
@owend @ljagiello if we are strongly opposed to doing this in vignette, I'm happy to turn this into a `HEAD` that provides the surrogate key in the headers. We can then wait until the new purging infrastructure is ready and use the `HEAD` to get the surrogate key for fastly purging.

macbre:
@drsnyder: Perl purger (aka the old one) does support surrogate keys - this feature is currently used by Interactive Maps only.

nmonterroso:
Closing, since this was solved a different way

PR #72, opened 2015-02-23
MAIN-3931 Do not add the format to the filename if thumb extension is the same as the original

drsnyder:
Do not add the format to the options if the thumbnail extension is the same as the original. This should result in a closer mapping between legacy file names and new file names. 

There is still more work to be done to make it easier to test this, but this should at least eliminate the unnecessary addition of `[format=foo]` into the legacy filenames when no format changes are taking place.

https://wikia-inc.atlassian.net/browse/MAIN-3931

/cc @ljagiello @nmonterroso


nmonterroso:
LGTM :shipit:

ljagiello:
awesome !

PR #73, opened 2015-02-24
MAIN-3930 Check the mime time to see if it can be thumbnailed earlier in the call stack

drsnyder:
This could prevent 2 round trips to CEPH by short circuiting the thumbnailing sooner. First check the name of the original before we go to CEPH and reject it if it's unsupported. Next, check again in the original location before we attempt to thumbnail the original (this is after the original has been downloaded).

This could catch supported files that happened to be mislabeled e.g. a JPEG labeled `.ogv`. In practice I doubt this happens that often. The opposite is also true but will be handled by the current implementation. That is there could be some OGG video files with the .jpg extension. We'll still catch these when the image is on disk locally.

https://wikia-inc.atlassian.net/browse/MAIN-3930

/cc @nmonterroso @ljagiello

### DIFF
--- a/src/vignette/util/thumbnail.clj
+++ b/src/vignette/util/thumbnail.clj
@@ -58,6 +58,7 @@
 
 (defn original->thumbnail
   [resource thumb-map]
+  (assert-original-mime-type resource thumb-map)
   (let [temp-file (temp-filename (str (wikia thumb-map) "_thumb"))
         base-command [thumbnail-bin
                       "--in" (.getAbsolutePath resource)
--- a/src/vignette/util/thumbnail.clj
+++ b/src/vignette/util/thumbnail.clj
@@ -37,7 +37,7 @@
 (defn assert-original-mime-type
   [file thumb-map]
   (let [mime-type (mime-type-of file)]
-    (when (contains? unsupported-mime-types (mime-type-of file))
+    (when (contains? unsupported-mime-types mime-type)
       (throw+ {:type :convert-error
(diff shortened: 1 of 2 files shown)


src/vignette/util/thumbnail.clj
```diff
@@ -79,6 +79,7 @@
 
 (defn get-or-generate-thumbnail
   [system thumb-map]
+  (assert-original-mime-type (get thumb-map :original) thumb-map)
```
nmonterroso: This just gets the mime type from a string? If so, it's relying on the file extension, which isn't always there. If these images pass this check, then I'd say LGTM

drsnyder: > If these images pass this check, then I'd say LGTM

By that do you mean does this catch `foo.ogv`? Yes, it does.

nmonterroso: I mean, does it allow `/a/ab/SomeImageWithoutAnExtension` through and try to thumbnail it. Weird edge case, but we do have a lot of images without extensions in CEPH

drsnyder: ` (u/assert-original-mime-type "file" {}) ` does not throw an exception. I should add a test for that.

drsnyder: I added tests in 43cf885. Ok to merge?

nmonterroso: LGTM :shipit: !

PR #74, opened 2015-02-25
Provide sane defaults for videos using mid-* in the thumbnail filename

drsnyder:
There are some legacy requests that attempt to thumbnail videos using `mid-<video>.ogv.jpg`. This enables sane handling of those requests with a broken image placeholder and a 404.

https://wikia-inc.atlassian.net/browse/PLATFORM-912

/cc @nmonterroso @ljagiello

### DIFF
--- a/src/vignette/api/legacy/routes.clj
+++ b/src/vignette/api/legacy/routes.clj
@@ -5,6 +5,8 @@
             [vignette.util.regex :refer :all])
   (:import [java.net URLDecoder]))
 
+(def default-width 200)
+
 (declare route->revision
          route->dimensions
          route->offset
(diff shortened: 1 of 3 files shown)


ljagiello:
:+1:

src/vignette/api/legacy/routes.clj
```diff
@@ -172,6 +174,10 @@
   "Add the :width field to a request map based on the legacy parsing methods."
   (if-let [thumb-dimension (:dimension map)]
     (cond-let
+      ; this matches some legacy video requests that attempt to take the width from a frame in the "middle"
```
nmonterroso: Oh, is that what it does? Was always a mystery to me.

nmonterroso:
LGTM

src/vignette/api/legacy/routes.clj
drsnyder: The magic can be found around [here](https://github.com/Wikia/backend/blob/master/lib/Wikia/Thumbnailer/Video.pm#L28).

PR #75, opened 2015-02-28
Consistent request map handling between old and new URI

drsnyder:
There were a few things that stood out to me as I was working on [the duplicate file issue (PLATFORM-940)](https://wikia-inc.atlassian.net/browse/PLATFORM-940). One was that the way we handle the generation of the `image-params` was significantly different between the legacy handling and the vignette handling. The second was that I was pretty sure we were generating other duplicate files other than `[format=ext]`.

This PR addresses both issues. First, the `image-params` map generation has been made more consistent. As a result it's now possible to test that a legacy URI and a new URI that we expect to generate the same thumbnail path in fact to so. See the tests that were added. As a result, this uncovered a difference in how the parameters were being computed for the window cropping in the new style URIs. That also has been fixed.

The `image-params` generation now uses the threading macro, `->` in both cases. I find that this form is a lot more readable and it makes it easier to test the new URIs.

This conflicts with #71 but I think it's ok. I'm much happier with this code and would rather build upon this work than that to add the `HEAD` support.

/cc @nmonterroso @owend

### DIFF
--- a/src/vignette/http/routes.clj
+++ b/src/vignette/http/routes.clj
@@ -117,7 +117,7 @@
         (GET window-crop-fixed-route
              request
              (handle-thumbnail system
-                               (route->thumbnail-auto-height-map
+                               (route->thumbnail-map
                                  (:route-params request)
                                  request)))
         (GET thumbnail-route
--- a/src/vignette/http/routes.clj
+++ b/src/vignette/http/routes.clj
@@ -86,6 +86,7 @@
          handle-original
          get-image-params
          route->offset
+         route->adjust-window-offsets
          route->thumbnail-map
          route->thumbnail-auto-height-map
          route->options
--- a/src/vignette/http/routes.clj
+++ b/src/vignette/http/routes.clj
@@ -82,10 +82,8 @@
                   :thumbnail-mode "scale-to-width"
                   :width size-regex}))
 
-(declare image-request-handler
-         handle-thumbnail
+(declare handle-thumbnail
          handle-original
-         get-image-params
          route->offset
          route->adjust-window-offsets
          route->thumbnail-map
@@ -186,15 +184,6 @@
       (request-timer)
       (add-headers)))
 
-; TODO: remove
-(defn image-request-handler
-  [system request-type request &{:keys [thumbnail-mode height] :or {thumbnail-mode nil height nil} :as params}]
-  (let [image-params (get-image-params request request-type)
-        image-params (if params (merge image-params params) image-params)]
-    (condp = request-type
-      :thumbnail (handle-thumbnail system image-params)
-      :original (handle-original system image-params))))
-
 (defn handle-thumbnail
   [system image-params]
   (if-let [thumb (u/get-or-generate-thumbnail system image-params)]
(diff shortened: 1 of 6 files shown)


ljagiello:
@drsnyder It is highly possible you're right and we're generating more duplicates. Recently I changed number of days we're keeping thumbs from 15 -> 10.

https://ganglia.wikia-inc.com/graph.php?r=month&z=xlarge&m=objects_in_cluster&h=dfs-s1&c=DFS&vl=object&title=objects+in+cluster

As you can see it's a higher number of deletes per day but same time total number of files is not going down.

src/vignette/http/routes.clj
```diff
@@ -87,39 +99,47 @@
...
+             (handle-thumbnail system
+                               (route->thumbnail-auto-height-map
+                                 (:route-params request)
+                                 request)))
         (GET window-crop-fixed-route
              request
-             (image-request-handler system :thumbnail request
-                                    :thumbnail-mode "window-crop-fixed"
-                                    :height :auto))
+             (handle-thumbnail system
```
nmonterroso: it looks like both `window-crop` and `window-crop-fixed` will generate the same thumbnail, since they both add `:height :auto` to the map. Only `window-crop` should add `:height :auto`, `window-crop-fixed` should use what's in the URL

src/vignette/http/routes.clj
```diff
@@ -203,3 +203,50 @@
     (clojure.string/replace (:image-type route-params)
                             #"^\/(.*)"
                             "$1")))
+
+(defn route->image-type
```
nmonterroso: Should we move all these `route->` functions to a different namespace?

nmonterroso:
Legacy URLs and Vignette URLs calculate window width/height differently because the parameters to the image mean different things. In the legacy format `a,b,c,d`, `a` is `x-offset`, `b` is `x-endpoint` (which is not specified at all in Vignette URLs), and `c,d` are the same but for `y`. In Vignette URLs, we specify the `x-offset`, which is the same as `a` above, but we specify `window-width` instead of `x-endpoint` (where `window-width` is `x-endpoint - x-offset`).

Some of the changes in this PR seem to be assuming that `a,b,c,d` maps directly to `x-offset`, `x-width`, `y-offset`, and `y-width`.

src/vignette/http/routes.clj
drsnyder: Good catch. I think this should be `route->thumbnail-map` since the `:thumbnail-mode` should be in the `image-params` along with everything else that's in the URL. Fixed in b7dfbfa.

src/vignette/http/routes.clj
drsnyder: Yeah. I was going to save that for later if it's ok. We have the same issue in the legacy routes namespace.

nmonterroso: :+1:

drsnyder:
> Legacy URLs and Vignette URLs calculate window width/height differently because the parameters to the image mean different things. In the legacy format a,b,c,d, a is x-offset, b is x-endpoint (which is not specified at all in Vignette URLs), and c,d are the same but for y.

Ah ha. We should document that somewhere. Since that is the case, it's very unlikely that we are generating duplicates for these kinds of requests.

nmonterroso:
LGTM :shipit:

PR #76, opened 2015-03-05
Only require --height when necessary

drsnyder:
We are currently requiring the --height param even though some of the thumbnailing modes do not require it. I'm pulling this over from #71 since that PR is going to be ditched in favor of adding `HEAD`.

/cc @nmonterroso


nmonterroso:
LGTM

PR #77, opened 2015-03-05
Refactor routes namespaces to move the helper functions out

drsnyder:
This is a follow-up to the feedback [here](https://github.com/Wikia/vignette/pull/75#discussion_r25633018).

This PR attempts to simplify our routing namespaces by moving the helper functions out into a separate namespace. This should reduce the clutter and make the main `http.routes` namespace more cohesive.

I also started an experiment with a very simple integration test to see if it works in travis. If it does we can add more to make sure that the API remains stable and ensure that the correct headers are in place. I think our test coverage is decent, but insufficient when making changes that affect the outermost part of the request stack like the handler as in the case with this PR.

/cc @nmonterroso

### DIFF
--- a/.midje.clj
+++ b/.midje.clj
@@ -1,0 +1,1 @@
+(change-defaults :fact-filter (complement :thumbnail-integration))
--- a/project.clj
+++ b/project.clj
@@ -10,20 +10,20 @@
                  [clout "1.2.0"]
                  [compojure "1.3.1"]
                  [com.novemberain/pantomime "2.3.0"]
+                 [digest "1.4.4"]
                  [environ "0.5.0"]
                  [info.sunng/ring-jetty9-adapter "0.8.1"]
                  [ring "1.3.2"]
                  [slingshot "0.10.3"]
                  [useful "0.8.8"]
-                 [digest "1.4.4"]
                  [wikia/commons "0.1.3-SNAPSHOT"]]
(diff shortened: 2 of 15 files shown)


drsnyder:
Duh, This would never work in travis. I'll make it so that you have to supply `:integration` explicitly to the command line to run the integration tests.

project.clj
```diff
@@ -23,6 +23,7 @@
                                      [javax.servlet/servlet-api "2.5"]
                                      [org.clojure/tools.namespace "0.2.5"]
                                      [org.clojure/tools.trace "0.7.8"]
+                                     [clj-http "1.0.1"]
```
nmonterroso: I've been trying so hard to keep these things in alphabetical order!

.midje.clj
```diff
@@ -0,0 +1 @@
+(change-defaults :fact-filter (complement :thumbnail-integration))
```
nmonterroso: wat is this

nmonterroso:
LGTM

.midje.clj
drsnyder: It's midje's configuration for fact filtering. Is says that the default should not include that key. This is so those integration tests won't run in Travis.

project.clj
drsnyder: Ok, f34d0da.

PR #78, opened 2015-03-09
PLATFORM-826 Add head support & integration tests for vignette routes

drsnyder:
This PR adds `HEAD` support to vignette routes to satisfy the vignette side the new purger integration. This turned out to be a little more complicated than expected and the behavior depends upon the order of the route definition. Specifically `HEAD` needs to be defined before `GET` because of [this block in compojure](https://github.com/weavejester/compojure/blob/639fffd3f8b77153644d0c9b87bca5af56be0cb0/src/compojure/core.clj#L28).

One way to handle `HEAD` is to set the `:body` to `nil` in a `GET` request. It looks like this is a shortcut that is imposed by compojure. This is probably fine in the case of materialized bodies but could be problematic when working with file streams as we are.

This might be an pre-mature optimization but in our case, we probably want to make sure that we don't open any more connections than necessary. To this end, the `HEAD` support provided here is pragmatic in that it only tests to see if the object exists in CEPH. It also means that the response does not include the `Content-Type` or the `Content-Length`. Omitting the latter is probably expected, but probably not the former.

Also of note is that the `Surrogate-Key` header will now be supplied when the resource is not found (e.g. 404). This is because in practice, by the time we do the purge from the new purger the object will like have already been deleted in CEPH. If we omit the `Surrogate-Key` on 404 then we can't complete the purge.

Also note that legacy HEAD support is still only partially supported. The request will probably hang because the `Content-Length` will still be there because of the above noted issue with compojure.

https://wikia-inc.atlassian.net/browse/PLATFORM-826

/cc @nmonterroso @owend @ljagiello 

Todo:
  * [x] Update the documentation regarding HEAD

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ A Clojure library for thumbnail generation and storage.
             - [zoom-crop](#zoom-crop)
             - [zoom-crop-down](#zoom-crop-down)
         - [Note Regarding Legacy Compatibility](#note-regarding-legacy-compatibility)
+    - [Other HTTP Methods Supported](#other-http-methods-supported)
+        - [HEAD](#head)
 - [License](#license)
 - [Contributors](#contributors)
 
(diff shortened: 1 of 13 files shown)


nmonterroso:
Should the http code for `HEAD` requests be 204? Other than that LGTM

drsnyder:
According to the [RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.1) a `HEAD` should include the same headers as a `GET` including the status code. 

I think the 204 is for the case when a resource has no representation and only headers.

nmonterroso:
Ah OK. LGTM then. PR tradesies?

PR #80, opened 2015-03-09
prevent double url-decoding

nmonterroso:
@drsnyder @owend 
https://wikia-inc.atlassian.net/browse/PLATFORM-993

Our current versions of [compojure (1.3.1)](https://github.com/weavejester/compojure/blob/1.3.1/src/compojure/core.clj#L33) and [clout (1.2.0)](https://github.com/weavejester/clout/blob/1.2.0/src/clout/core.clj#L104) each url decode the request `path`, so when a request comes in that has `%25_something_else`, the `%25` is replaced by `%` by clout, and then `%_something_else` is attempted to be replaced by compojure. This can either lead to a 500 error when `_something_else` is not a number, or a 404 when the resulting url decoded name does not live in CEPH. Upgrading compojure and clout to 1.3.2 and 2.1.0, respectively, resolves this conflict. Compojure is now responsible for url decoding.


drsnyder:
Do you know what changed between the two to fix this issue?

nmonterroso:
Clout no longer url decodes. Compojure does.

drsnyder:
So before they were both url decoding? At these versions and above only compojure URL decodes?

nmonterroso:
In the past versions, yes. They were both url decoding. In the current versions, only compojure url decors.

test/vignette/http/legacy/routes_test.clj
```diff
@@ -243,7 +243,7 @@
                              (request :get "/intmap_tile_set_4823/thumb/20150205173220%21phpZDfa00.jpg/1110x300x5-20150205173220%21phpZDfa00.jpg")))]
     (:width map) => "1110"
     (:height map) => "300"
-    (:original map) => "20150205173220!phpZDfa00.jpg"
+    (:original map) => "20150205173220%21phpZDfa00.jpg"
```
drsnyder: This tests that clout is not getting in the way, but how do we know that compojure will do the right thing? Is there a way to test to make sure that compojure will url decode?

nmonterroso: I thought about adding tests to make sure that we eventually end up with a proper url-decoded path, but I don't know if it makes sense for us to unit test compojure. It has a [good amount of those](https://github.com/weavejester/compojure/tree/master/test/compojure).

drsnyder: I guess it's more of a functional test. We want to know that compojure does the right thing given an input. It may be more hassle then it's worth though because the interfaces might change.

drsnyder:
lgtm

test/vignette/http/legacy/routes_test.clj
nmonterroso: Yah, I was looking at where this is done and it looks like one of the middleware functions you get by default

PR #81, opened 2015-03-23
replace multiple slashes in :uri with a single slash

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-896

Adds middleware to replace multiple slashes with a single slash. These account for ~5% of our `bad-request-path` errors.

### DIFF
--- a/src/vignette/http/middleware.clj
+++ b/src/vignette/http/middleware.clj
@@ -1,6 +1,7 @@
 (ns vignette.http.middleware
   (:require [environ.core :refer [env]]
             [compojure.response :refer [render]]
+            [clojure.string :as string]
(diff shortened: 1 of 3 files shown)


drsnyder:
Test?

nmonterroso:
asking for tests is not how PR tradesies works :(

drsnyder:
> asking for tests is not how PR tradesies works :(

Shame on you for expecting a tradesie without tests :). BTW, we seem to be missing a guideline about PR tradsies. I think it would be a nice supplement to the existing [guidelines](https://github.com/Wikia/guidelines/blob/master/Process/PullRequests.md).

drsnyder:
:eyes: :+1: TM

PR #82, opened 2015-04-06
commas in quotes in content-disposition header

nmonterroso:
@drsnyder 
https://wikia-inc.atlassian.net/browse/PLATFORM-1053

Looks like a `Content-Disposition` header with quotes containing a comma (ex: `Content-Disposition: inline; filename="some_filename\"with,stuff\""`) cause a duplicate header problem in Chrome. Trick seems to be to return a literal `\"` back, which means escaping the quote to be `\\\"` :/

### DIFF
--- a/src/vignette/util/image_response.clj
+++ b/src/vignette/util/image_response.clj
@@ -1,5 +1,6 @@
 (ns vignette.util.image-response
   (:require [clojure.java.io :refer [file]]
+            [clojure.string :as string]
             [compojure.route :refer [not-found]]
             [ring.util.response :refer [response status header]]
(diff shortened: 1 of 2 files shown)


nmonterroso:
Note that in Chrome, the filename will be `some_filenamewith,stuff` (quotes completely removed) and in FireFox it will be `some_filename"with,stuff"` (quotes maintained, as expected)

drsnyder:
It feels like we are patching and patching this feature. RFC support looks like it's [non-trivial](http://greenbytes.de/tech/tc2231/). 

LGTMFN

PR #83, opened 2015-04-07
call force-thumb? for Vignette-style urls

nmonterroso:
https://wikia-inc.atlassian.net/browse/PLATFORM-1121
@drsnyder 

Looks like the functionality to force a thumbnail is only happening for legacy URLs. I also made sure the `X-Vignette-Force-Thumb` header is being properly created by Varnish.


drsnyder:
lgtm

PR #84, opened 2015-04-09
PLATFORM-966: cache images for 14 days

macbre:
Read the full story in https://wikia-inc.atlassian.net/browse/PLATFORM-966

Cache images for 14 days on both CDN and client level

With `s-maxage` being set to a value higher than `max-age` we were
getting lots of requests from clients trying to revalidate the image
because `Age` header (returned by CDN) was higher than `max-age` sent to
the client.

14 days will cover ~75% of images:

```
Age 5th percentile - 1.10 days
Age 10th percentile - 1.85 days
Age 25th percentile - 3.86 days
Age 50th percentile - 9.94 days
Age 75th percentile - 15.57 days
Age 90th percentile - 31.94 days
Age 95th percentile - 48.79 days
Age 99th percentile - 50.78 days
```

@nmonterroso / @drsnyder / @ljagiello

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ pom.xml.asc
 /tmp
 *.iml
 /logs
+*.swp
--- a/src/vignette/http/middleware.clj
+++ b/src/vignette/http/middleware.clj
@@ -58,9 +58,8 @@
   (let [status (get response :status 0)]
     (cond
       (and (>= status 200) (< status 300))
-      (header response cache-control-header (format "public, s-maxage=%d, max-age=%d",
-                                                    (hours-to-seconds (* 24 365))
-                                                    (hours-to-seconds 24)))
+      (header response cache-control-header (format "public, max-age=%d",
+                                                    (hours-to-seconds (* 14 24))))
 
       (and (>= status 400) (< status 500))
       (header response cache-control-header (format "public, max-age=%d"
--- a/src/vignette/http/middleware.clj
+++ b/src/vignette/http/middleware.clj
@@ -59,7 +59,7 @@
     (cond
       (and (>= status 200) (< status 300))
       (header response cache-control-header (format "public, max-age=%d",
-                                                    (hours-to-seconds (* 14 24))))
+                                                    (hours-to-seconds (* 365 24))))
 
       (and (>= status 400) (< status 500))
       (header response cache-control-header (format "public, max-age=%d"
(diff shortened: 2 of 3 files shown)


drsnyder:
> Age header (returned by CDN) was higher than max-age sent to
the client

How do we know this is happening?

What are our alternatives for handling this issue? Could/should we consolidate `max-age` on 1 year?

macbre:
@drsnyder - it does happen, take a look on `curl` examples in the ticket linked above.

We can set `max-age` to 1 year (and skip `s-maxage`), but this will create problems with deleted images being kept in browser caches. I do not know how big problem that is. Old thumbnailer used to set `max-age` to one year since its very beginnings and I do not recall any issues related to that.

@ljagiello: one year sounds fine?

macbre:
@drsnyder - my concern with deleted images comes from your comment to a3df36a

drsnyder:
@macbre what if we ease into this by raising `max-age` to 14 days and leaving `s-maxage` as-is and see what happens to bandwidth? Ideally I think we want to cache images in fastly as long as possible.

macbre:
I don't think that increasing `max-age` and keeping `s-maxage` significantly larger will make any change. The most important question here is: why do we want to cache shorter on the client side than on CDN?

drsnyder:
> why do we want to cache shorter on the client side than on CDN?

Primarily because we have no control over the client. But if you don't think that's an issue then let's just get rid of `s-maxage` and set them both to 1 year. I'll trust your judgement here.

ljagiello:
Macbre check :)
Macbre_check

macbre:
@drsnyder: will change to a year.

Thanks guys!

macbre:
@drsnyder / @nmonterroso: please review the changes. I'll deploy these changes on Monday.

drsnyder:
lgtm

macbre:
Thx

macbre:
Is dt prep / push for vignette app all that's required to deploy these changes?

drsnyder:
I normally:

```
dt prep -a vignette -e dev-vignette -r vignette@master
dt push -a vignette -e dev-vignette
# quick sanity check in dev
dt prep -a vignette -e prod -r vignette
dt push -a vignette -e prod
```

macbre:
Thanks! I'll do my best not to break anything ;)

macbre:
Deployed

PR #87, opened 2015-06-03
PLATFORM-1236: implement scale-only-down

michalroszka:
The JIRA case: https://wikia-inc.atlassian.net/browse/PLATFORM-1236

## Changes ##

1. scale-to-width-down flow added to `bin/thumbnail` — if the requested image is bigger on either side than the original one, keep the original dimensions
2. scale-to-width-down route added to `src/vignette/http/routes.clj` — just a route to the above
3. scale-to-width-down renamed to scale-only-down
4. scale-only-down does not require width and can base on height, too
5. scale-only-down keeps the original dimensions when no height nor width has been given
6. scale-only-down removed, replaced with separate scale-to-width-down and scale-to-height-down
7. thumbnail-auto-width request handler created for scale-to-height-down
8. correspondig routes created

## To Do ##

1. tests in `test/vignette/http/routes_test.clj`
2. tests in `test/vignette/http/integration_test.clj`
3. tests in `test/vignette/media_types_test.clj`

## Code Reviewers ##

@nmonterroso @drsnyder 

## Stakeholders ##

@RafLeszczynski @jcellary


bin/thumbnail
```diff
@@ -2,7 +2,7 @@
 # Wrapper for imagemagick that encapsulates our thumbnail logic.
 # @author Damon Snyder (damon@huddler-inc.com) 
 
-IMAGEMAGICK_BASE=${IMAGEMAGICK_BASE:-"/usr/local/"}
```
michalroszka: @drsnyder @nmonterroso — the trailing slash seems unnecessary as the variable is only used to define `CONVERT` and `IDENTIFY` below.

nmonterroso: :+1:

src/vignette/http/routes.clj
```diff
@@ -76,6 +76,17 @@
                   :thumbnail-mode "scale-to-width"
                   :width size-regex}))
 
+(def scale-only-down-route
```
nmonterroso: Looks like this route isn't getting used anywhere, it needs to be added to [`app-routes`](https://github.com/Wikia/vignette/blob/platform-1236-implement-scale-to-width-down/src/vignette/http/routes.clj#L107)

bin/thumbnail
```diff
@@ -313,6 +318,17 @@ function thumb_params() {
 		"scale-to-width")
 			THUMB_PARAMS="${QUALITY} -thumbnail ${WIDTH}"
 			;;
+		"scale-only-down")
+			# get height and width
+			origheight=$($IDENTIFY -format "%h\n" "$IN" | head -1)
+			origwidth=$($IDENTIFY -format "%w\n" "$IN" | head -1)
+
+			if [[ "$WIDTH$HEIGHT"X = "X" || $HEIGHT -gt $origheight || $WIDTH -gt $origwidth ]]; then # requested is bigger on either side, so we keep the original dimensions
```
nmonterroso: i think `$HEIGHT` here will be empty, assuming the code flows similarly to `scale-to-width`

bin/thumbnail
```diff
@@ -38,6 +38,7 @@ MODE_OPTIONS=(
 	"fixed-aspect-ratio"
 	"fixed-aspect-ratio-down"
 	"scale-to-width"
+	"scale-only-down"
```
drsnyder: For consistency, can we call this `scale-to-width-down`?

michalroszka: Technically, it is "scale down to width or height, whichever given". Any other naming proposal? What about simply "scale down"?

src/vignette/http/routes.clj
michalroszka: I haven't firgured it out yet. Thanks for the pointers. :)

bin/thumbnail
michalroszka: `width` and `height` are both optional. When none if them is provided, we keep the original size. When one of them is given, we respect the one. When both are provided, we respect both. I'll fix that in the code, thanks.

src/vignette/http/routes.clj
```diff
@@ -76,6 +76,28 @@
...
+
+(def scale-to-height-down-route
+  (route-compile "/:wikia:image-type/:top-dir/:middle-dir/:original/revision/:revision/:thumbnail-mode/:height"
+                 {:wikia wikia-regex
+                  :image-type image-type-regex
+                  :top-dir top-dir-regex
+                  :middle-dir middle-dir-regex
+                  :original original-regex
+                  :revision revision-regex
+                  :thumbnail-mode "scale-to-width-down"
```
drsnyder: I think this should be `scale-to-height-down` (`s/width/height/`).

drsnyder:
Please add tests for the new routes [here](https://github.com/Wikia/vignette/blob/master/test/vignette/http/routes_test.clj) and in the [integration tests](https://github.com/Wikia/vignette/blob/master/test/vignette/http/integration_test.clj) as well.

You have to run the integration tests manually since they are not part of the travis build. You might also need to install imagemagick. If you are running Linux that should be pretty easy. 

To run the integration tests, make sure you setup your environment. See the README. The main setting you'll need is `VIGNETTE_THUMBNAIL_BIN=$PWD/bin/thumbnail`.

Once that is set, run `lein midje :thumbnail-integration`. To add new integration tests, I typically curl localhost for the new URL I'm adding then md5 the file that it pulled down. I'll then add that to the test to make sure I got the body I expected.

If you want to run the server at the repl, run 

```
$ lein repl
...
user=> (create-integration-env)
user=> (start system-local 8888)
```

Then you can curl localhost on port 8888 with the new paths. See the integration tests for some sample image file names you can use for testing.

michalroszka:
Thanks @drsnyder — as mentioned in the To Do section, writing tests is my next step :) I needed to figure out some basics about deployment and manual testing on `dev-vignette`, documented here https://one.wikia-inc.com/wiki/Platform_Team/Vignette#Contributing_to_Vignette

src/vignette/http/routes.clj
michalroszka: This one is for `width`. The one for `height` is below.

drsnyder: Sorry, my eyes must have failed me. It must be the hair loss.

michalroszka: Covering the eyes with _the_ hair is so difficult these days, yeah.

michalroszka:
I can't run the integration tests:
```
 ~/Desktop/Wikia/vignette cat script.sh 
#!/bin/bash
export LOGGER_APPLICATION=vignette
export LOGGER_TYPE=file
export LOGGER_FILE_OUTPUT=/tmp/vignette.log
export VIGNETTE_TEMP_FILE_LOCATION=/tmp/vignette
export VIGNETTE_THUMBNAIL_BIN=$PWD/bin/thumbnail
export VIGNETTE_INTEGRATION_ROOT=/tmp/integration
export IMAGEMAGICK_BASE=/usr/bin
```

michalroszka:
I have fixed the log4j warnings:
```
 ~/Desktop/Wikia/vignette cat src/log4j.properties 
log4j.rootLogger=INFO, stdout

log4j.appender.stdout=org.apache.log4j.ConsoleAppender
log4j.appender.stdout.Target=System.out
log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
log4j.appender.stdout.layout.ConversionPattern=%d{ABSOLUTE} %5p %c{1}:%L - %m%n
```

michalroszka:
I am also getting this:
```
user=> (create-integration-env)

CompilerException java.lang.RuntimeException: Unable to resolve symbol: create-integration-env in this context, compiling:(/tmp/form-init3871826680541476243.clj:1:1)
```

michalroszka:
Oh, and here is a new problem:
```
(Could not find metadata joda-time:joda-time/maven-metadata.xml in local (/home/michal/.m2/repository))
(Could not transfer metadata joda-time:joda-time/maven-metadata.xml from/to central (https://repo1.maven.org/maven2/): Checksum validation failed, expected bed66f1e4d165a744fb45fe60a4e0afbfa12cc78 but is 458b940ded0c9495327658ff0cc1dfb13b3bc324)
This could be due to a typo in :dependencies or network issues.
If you are behind a proxy, try setting the 'http_proxy' environment variable.
```
@drsnyder - is there anything else I should know about setting up the environment?

drsnyder:
Hi @michalroszka sorry for the delayed response. It looks like we don't have any overlapping hours. Normally, the 500 error is when there is something wrong with the environment. For the integration errors, try the following:

 * make sure you have imagemagick installed. To test, see if you can run `convert` and `$IMAGEMAGICK_BASE/bin/convert`
 * set `export CONVERT_CONSTRAINTS=" "` in your setup script
 * can you run `$VIGNETTE_THUMBNAIL_BIN --in image-samples/boat.jpg --out /tmp/foo.jpg --width 100 --mode scale-to-width`? Does it create /tmp/foo.jpg? You can also use this to test your new thumbnail mode.
 * if the above doesn't work, cat `/tmp/vignette.log` after you try and run the integration test


For `(create-integration-env)`, sorry, this should be ` (itg/create-integration-env)`. You have to add the prefix because in `dev/user.clj` the integration namespace is required as `[vignette.util.integration :as itg]` which means you have to prefix the symbols in that namespace with `itg/`.

drsnyder:
For the metadata issue-- I've never seen that before. The closing thing that google provided was [this](http://stackoverflow.com/questions/7819163/maven-failed-to-retrieve-plugin-descriptor-error) from SO. That sounds like a connectivity or proxy issue. But if that was the case I would have expected errors earlier.

drsnyder:
@michalroszka sorry, I found the problem. This worked for me. The key part is `export IMAGEMAGICK_BASE=/usr/`.

```
[damon@dev-damon vignette]$ export VIGNETTE_THUMBNAIL_BIN=$PWD/bin/thumbnail
[damon@dev-damon vignette]$ export IMAGEMAGICK_BASE=/usr/
[damon@dev-damon vignette]$ lein repl                                                                                                                              
(WARNING!!! version ranges found for:)
([clj-aws-s3 "0.3.10"] -> [com.amazonaws/aws-java-sdk "1.7.5"] -> [joda-time "[2.2,)"]
Consider using [clj-aws-s3 "0.3.10" :exclusions [joda-time]].)
nil
2015-06-15 16:00:47.630:INFO::main: Logging initialized @15941ms
nREPL server started on port 49924 on host 127.0.0.1 - nrepl://127.0.0.1:49924
REPL-y 0.3.2, nREPL 0.2.3
Clojure 1.5.1
Java HotSpot(TM) 64-Bit Server VM 1.8.0_31-b13
    Docs: (doc function-name-here)
          (find-doc "part-of-name-here")
  Source: (source function-name-here)
 Javadoc: (javadoc java-object-or-class-here)
    Exit: Control+D or (exit) or (quit)
 Results: Stored in vars *1, *2, *3, an exception in *e

user=> (load-facts 'vignette.http.integration-test :thumbnail-integration)
```

This did reveal a problem. The integration tests (that check the file sizes and hashes) probably only work on a specific version of image magick on a specific architecture. I wouldn't worry about fixing those for now. We can create a ticket to target them for a box that has vignette installed via chef.

drsnyder:
As I said, I wouldn't worry about the integration tests. If you can:

```
(itg/create-integration-env)
(start system-local 8888)
```

Then render some of the sample images using the new modes we can :ship: it.

michalroszka:
I did checked my changes by running `thumbnail` script in the command line while developing the feature. I have also checked it the routing  on `dev-vignette`.

michalroszka:
Thanks for the help, @drsnyder! I have managed to get it working. I will merge the changes and release vignette. The only remaining problem is the integration tests. We need to run them in unified environment which creates some technical debt with regard to future updates. But this can be automated, too.

drsnyder:
> The only remaining problem is the integration tests. We need to run them in unified environment which creates some technical debt with regard to future updates. But this can be automated, too.

Yeah, Nelson and I have been talking about this too. There's probably a ticket floating around for it.

PR #88, opened 2015-06-09
Fix the integration tests

drsnyder:
* Fixed the port
 * Fixed the max-age settings

/cc @nmonterroso @michalroszka

### DIFF
--- a/test/vignette/http/integration_test.clj
+++ b/test/vignette/http/integration_test.clj
@@ -8,6 +8,9 @@
             [vignette.storage.core :refer [create-image-storage]]
             [vignette.util.integration :refer [create-integration-env integration-path]]))
 
+; If you get 500 errors it's probably because your environment is not setup. See the README to make sure
+; the thumbnail script location is set.
+
 (create-integration-env)
 
 (def default-port 8888)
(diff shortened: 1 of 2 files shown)


test/vignette/http/integration_test.clj
```diff
@@ -16,97 +19,97 @@
 (def system-local (create-system lis))
 
 (with-state-changes [(before :facts (start system-local default-port))
-                     (after :facts (stop system-local))]
+                     (after :facts (do (stop system-local)))]
```
michalroszka: Not really code review, just a question. Does the `do` here that the results of `stop system-local` will be returned "up to the teardown call", rather than just executed?

michalroszka:
I just have one question about the `do`. All other changes are literal and correct. :)

test/vignette/http/integration_test.clj
drsnyder: The `do` was actually a mistake. I didn't mean to commit that. Typically `do` is used when you want to group multiple expressions that have side effects. Something like:

```
(do
  (execute-something)
  (stop system-local))
```

drsnyder: Removed in a2a3e27.

PR #90, opened 2015-07-02
Add a toggle for caching thumbnails in vignette

drsnyder:
Added `:cache-thumbnails` to storage core and the command line interface. When this toggle is true, thumbnails will not be read from nor written to the backing storage. They will always be generated at request time.

https://wikia-inc.atlassian.net/browse/SERVICES-367

/cc @nmonterroso @ljagiello

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,19 @@ Below is a list of environment variables that will affect the vignette runtime.
  * `CONVERT_CONSTRAINTS`          universal options to pass to ImageMagick. see bin/thumbnail
  * `UNSUPPORTED_REDIRECT_HOST`    on an unsupported legacy thumbnail request, host to redirect
 
+## Command Line Options
+
+To see the command line options available execute `lein run -- -h`. The notable
+command line options are the following:
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -69,6 +69,19 @@ Below is a list of environment variables that will affect the vignette runtime.
...
  * `UNSUPPORTED_REDIRECT_HOST`    on an unsupported legacy thumbnail request, host to redirect
 
+## Command Line Options
+
+To see the command line options available execute `lein run -- -h`. The notable
+command line options are the following:
+
+ * -C,--cache-thumbnails: enable thumbnail caching. When this option is provided
+   thumbnails will be written to and read from the backing storage provided. The
+   default is false.
```
nmonterroso: From the code below it looks like the default is true

nmonterroso: Err, sorry no but the default is not specified in the `cli-specs` - does an absent CLI option default to `false`?

nmonterroso:
question about CLI but OTW LGTM

README.md
drsnyder: The default should be `nil` which is falsey. To make it clear, I'll add `:default false`.

PR #91, opened 2015-07-02
Add counters for thumbnail generation

drsnyder:
Add counter for thumbnail generation and thumbnail cache hits. This will give us more insight into how often we are using the backing cache for thumbnails.

/cc @ljagiello @nmonterroso


nmonterroso:
LGTM

PR #94, opened 2015-07-23
Switch sunng87/ring-jetty9-adapter to mpenet/jet and upgrade to jetty 9.3.0

pchojnacki:
Additionally switch stdout logging to log to Syslog/LOCAL0

----
According to profiler fixing logging for Jetty and S3 reduced time Vignette was spending processing  logs. Which contributed small percent to performance improvement this PR brings.

Additionally when researching options for async communication I've found that just upgrading to jetty 0.9.3 (via mpenet/jet)  improves the throughput of Vignette on my laptop by around 10% i.e. from 80-90 rps to 90-100 rps. Also reducing Memory usage and CPU. Also threads are now blocking much less and doing more work.

This might be due to Jetty switching to async processing internally in version 9.3.0 
https://webtide.com/jetty-9-3-features
 
/cc: @nmonterroso @drsnyder @ljagiello @cmiller-wikia 

I also tested other frameworks that provide async communication namely aleph and http-kit but it seems that they don't provide this kind of improvement out of the box and are also incompatible with few our modifications to logging.


pchojnacki:
Small clarification:
I've tested it using locally installed Ceph/S3 and not Local mode

ljagiello:
@pchojnacki What value for `STORAGE_MAX_CONNS` you want to use at production? Default?

drsnyder:
Hi @pchojnacki! Was this PR prompted by a specific issue? I don't want to ignore performance but we also have a fairly long backlog of image processing issues in Jira.

pchojnacki:
@ljagiello When testing with 100 concurrent connections I've bumped to the default limit of 50 and got errors that Vignette timed out when waiting for free worker. So for my testing I decided to use the same number for max-conns as the number of Jetty worker threads that process incoming connections so in case of cache fetches there is 1:1 relation of workers. 

So the factory default is actually 50 but I'd suggest keeping the pool size at around 150. Also since I wasn't able to find this specific problem happening on production I'm open to suggestions.

pchojnacki:
@drsnyder I was doing some investigation into async frameworks for purposes of integration with Static Assets service and I noticed different patterns in threads in VisualVM. Since its hackathon and I also noticed the ongoing thread about Thumbnail caching I decided to investigate a little bit more.

drsnyder:
> Since its hackathon and I also noticed the ongoing thread about Thumbnail caching I decided to investigate a little bit more.

FWIW, I think we want more data about where the trouble was coming from. If you wanted to measure that, just remove the `-C` command line option for vignette on one of the thumbnailer machines and attach visualvm to it. That will give you some production data to investigate where we might be able to reduce the CPU usage. visualvm might no tell you the whole story though because a large part of the CPU usage was coming from imagemagick.

pchojnacki:
I decided to tackle this part first since I remember talking to Nelson some time ago and that some of the problems Vignette used to have was not entirely related to ImageMagick. Although I might have remembered wrong :D

If I won't be doing anything else tomorrow I might look into it too. 
I used to be sure that some performance gaps of vignette might be coming not from ImageMagick itself but from Bash and the amount of processess it is spawning to handle the request so I might try to isolate this and see if IMagick+Bash combo will be able to process more thumbnailing request than some native Java library.  Because otherwise as I did some benchmarks in our old email discussion like a year ago those benchmarks showed that Imagick is quite fast at processing images and that probably dominates the "heatmap" of cpu usage but maybe bash makes it sufficiently slower so that other solutions will be faster.

This PR is more about trying to make Vignette as close to Varnish in pure data shuffling throughput as possible while also adding support for async operations so if we will stick to some form of persistent/semi-persistent thumbnail cache the cache lookups will not need dedicated thread pool.

drsnyder:
> So the factory default is actually 50 but I'd suggest keeping the pool size at around 150. Also since I wasn't able to find this specific problem happening on production I'm open to suggestions.

This could in fact be a problem in production. If you tail the logs you'll see timeouts connecting to CEPH. I was suspicious about this but never had a chance to investigate. 

@pchojnacki would you mind splitting out the `STORAGE_MAX_CONS` change? I'm really curious to see if that has an impact on the error rate in production. It would be nice to separate the effect from the jetty changes.

pchojnacki:
The problem with small pool actually manifests as:
```
Unable to execute HTTP request: Timeout waiting for connection from pool : /bwtest/1/27/mario.png/revision/latest
```
I still can split the PR if you want but although I checked only on one host I haven't found any symptom of this problem. Since I think we average well below 50 rps on each dfs.
Although if we would increase the the timeout so that current gets will timeout less then I think this problem would have higher chances of occurring

drsnyder:
It looks like the latest version of [`ring-jetty9-adaper`](https://github.com/sunng87/ring-jetty9-adapter/blob/master/project.clj#L1) uses jetty 0.9.3. Should we try upgrading first to see if we see reduced CPU usage?

pchojnacki:
thanks for pointing https://github.com/sunng87/ring-jetty9-adapter/blob/master/project.clj#L1 I'll definetly take a look to try to confirm if the speedup is indeed caused by jetty internals change.

But I think I'll still stick with Jet because it provides some extension to Ring that makes it possible to use asynchronous requests processing directly. Which https://github.com/sunng87/ring-jetty9-adapter doesn't seem to provide.

drsnyder:
> But I think I'll still stick with Jet because it provides some extension to Ring that makes it possible to use asynchronous requests processing directly.

I would prefer not to make the change unless we need it. If we aren't currently using the asynchronous request processing extensions to Ring I don't think we should change.

pchojnacki:
Sorry for not including info earlier.
I'm working on using async communication to fetch images from Static assets service.

Its actually why I've upgraded that in the first place because when looking for async enabled servers I've tested a few and found that this one has some improvement out of the box. Using async enabled client will definitely improve performance for both thumbnail cache lookup as well as original downloads.

Additionally jet seems to be more popular at least in star-count and observed.

drsnyder:
> I'm working on using async communication to fetch images from Static assets service.

Ok, say more about that. Unless I'm mistaken that's a function of the HTTP client we use to fetch the objects from either CEPH or the SAS and not the request handler.

pchojnacki:
quoting from jet Readme:
>If you return a core.async channel in a ring body jetty will go into async mode and the channel values will be streamed as chunks. If the channel is closed the connection ends. If an error occurs or the client disconnects the channel closes as well.

This allows of handler code to also use async clients like the client bundled with Jet.

drsnyder:
I see now where this is headed. We were planning on doing this with `http-kit` but a while back we decided to consolidate on jetty since that's also what we were using with pandora. 

For the client side we'll need a new [storage handler](https://github.com/Wikia/vignette/blob/master/src/vignette/storage/s3.clj#L111-135). Are you planning on a followup PR for those changes?

PR #95, opened 2015-08-06
PLATFORM-1333 Vignette Static assets support

pchojnacki:
this change implements support for fetching images from Static Assets Service alongside usual images.
Additionally
* introduces some rudimentary support for finding services using consul.
* refactors routing that constitutes Vignette thumbnailing protocol so it can be reused within other contexts

Todo:
* [x] add tests to static_assets serving code
* [x] decouple service discovery from static asset


pchojnacki:
I'm still working on adding more tests and slight refactoring of ->ImageService but the my changes are mostly done.
Could you guys take a look ?
/cc : @nmonterroso and @drsnyder

drsnyder:
@pchojnacki please add a description to the PR.

src/vignette/http/routes.clj
```diff
@@ -115,26 +31,39 @@
...
+             :wikia wikia-regex
+             :image-type image-type-regex
+             :top-dir top-dir-regex
+             :middle-dir middle-dir-regex])
+
+(def uuid-context ["/:uuid"
+                   :uuid uuid-regex])
+
+(defmacro eval-context [ctx args routes]
+  `(context ~(eval ctx) ~args ~routes))
```
drsnyder: Is the `eval` necessary?

src/vignette/http/uuid/routes.clj
```diff
@@ -0,0 +1,60 @@
...
+
+(def scale-to-width-down-route
+  (route-compile "/:uuid/:thumbnail-mode/:width"
+                 {:uuid uuid-regex
+                  :thumbnail-mode "scale-to-width-down"
+                  :width size-regex}))
+
+(def scale-to-height-down-route
+  (route-compile "/:uuid/:thumbnail-mode/:height"
+                 {:uuid uuid-regex
```
drsnyder: The `:uuid` is duplicated here but the equivalent prefix is not also duplicated in the `proto` routes above. Is that intentional?

src/vignette/storage/core.clj
```diff
@@ -1,7 +1,8 @@
 (ns vignette.storage.core
   (:require [vignette.media-types :as mt]
             [vignette.storage.protocols :refer :all]
-            [vignette.util.query-options :as q]))
+            [vignette.util.query-options :as q]
+            [vignette.storage.static-assets :as sa]))
```
drsnyder: This namespace shouldn't need to know anything about a specific storage implementation.

project.clj
```diff
@@ -7,16 +7,18 @@
                  [org.clojure/tools.cli "0.3.1"]
                  [cheshire "5.3.1"]
                  [clj-aws-s3 "0.3.10"]
-                 [clout "2.1.0"]
-                 [compojure "1.3.2"]
+                 [compojure "1.4.0"]
+                 [consul-clojure "0.1.0-SNAPSHOT"]
```
drsnyder: I think we should avoid adding the consul dependency if at all possible. Can we use DNS to resolve the static asset service?

src/vignette/http/routes.clj
pchojnacki: yes, compojure expects to receive Vector of values this is the only way to achieve that vector will be 'inlined' and thus passed as a vector and not var or symbol.

src/vignette/http/uuid/routes.clj
pchojnacki: oh. thanks for spotting that Its some code I used before I refactored the main routes. Somehow I forgot to delete it

project.clj
pchojnacki: Why would like to avoid it ?
This here is only library that abstracts all the querying for me. DNS based querying will also be consul specific.

drsnyder: When we do this we are making `vignette` very wikia specific. We are coupling it to our service discovery choice and to the static asset service.

src/vignette/storage/static_assets.clj
```diff
@@ -0,0 +1,48 @@
...
+    (with-open [in-stream (io/input-stream (file-stream this))
+                out-stream (io/output-stream to)]
+      (io/copy in-stream out-stream))
+    (fs/file-exists? to))
+  (->response-object [this] (file-stream this)))
+
+(defn create-async-response-stored-object [oid]
+  (let [response @(http/get (build-url oid) {:as :stream})]
+    (if (-> response :status (= 200))
+      (->AsyncResponseStoredObject response))))
```
drsnyder: Is it possible to have vignette read from DFS directly? Alternatively, can we use an s3 client to read from the static asset service?

src/vignette/storage/static_assets.clj
```diff
@@ -0,0 +1,48 @@
+(ns vignette.storage.static-assets
+  (:require [vignette.storage.protocols :refer :all]
+            [org.httpkit.client :as http]
```
drsnyder: Previously I thought your intention was to use the [jet client](https://github.com/mpenet/jet#http-client). I don't know if the interfaces will be compatible between jet <-> httpkit.

src/vignette/storage/static_assets.clj
pchojnacki: Good questions we asked them ourselves when we started developing this. Its our intention to abstract that implementation detail away. Static assets service will be multi-dc/region agnostic the same cannot be said about s3. Also communicating with DFS directly violates service separatiton.

src/vignette/storage/static_assets.clj
pchojnacki: Jet client only supports pure async interface. While http-kit client provides the fallback to streaming interface as a thin wrapper on async callbacks. 
The idea is to compare the performance once jet client implements the streaming wrapper.
Alternatively the vignette might be refactored to core.async flow which would make jet client usable.

project.clj
pchojnacki: Good point I will work on abstracting the service discovery part so its less coupled.
But TBH our schema url  bucket/top/middle/revision/x is already very Wikia specific

src/vignette/storage/static_assets.clj
drsnyder: > Also communicating with DFS directly violates service separatiton.

Not if the static asset service provides an s3 interface. As it is now, vignette could be used on top of s3 or DFS (anything that speaks the s3 object protocol).

src/vignette/storage/static_assets.clj
drsnyder: > Jet client only supports pure async interface. 
This should work for the static asset service-- as long as it returns the core.async channel back to jet for the [body](https://github.com/Wikia/vignette/pull/95/files#diff-5f65fe56082f716f52998b3d0642f4d3R20) think it should would work seamlessly.

drsnyder: It could also work when reading from CEPH. We don't really use any of the s3 specific features. As far as I know it's just a `GET` on the object.

drsnyder:
@pchojnacki I wonder if another way to approach this is to move the `context` and the storage client into a clojure based configuration file. Something like the following:

```
; macro that creates the context and associates the storage object with it
(defcontex
   uuid-context ["/:uuid" :uuid uuid-regex] (create-image-storage sasos))
```

When a `request` map has all the keys of the context (`(rest context-mapping)`) then you use the storage object associated with it. If none map then you 404.

I think long term we want to switch to the `:uuid` context so we can deprecate the rainbow directories used by MW. I think it would be a lot of work to do that but it would be a much more generalized solution. The "storage" layer that you read from just has be simple and respond to `GET`. 

This might assume that long term we move away from thumbnail caching. I still think we haven't explored that option to it's fullest. We had one (very short) failed trial with no post mortem and it was written off as a "spectacular failure."

If we did that it would probably make vignette more generic and less MW or Wikia specific.

src/vignette/storage/static_assets.clj
ljagiello: Please use only S3 interface, supporting multiple different ways of access will only complicate stuff sooner or later.

src/vignette/util/consul.clj
```diff
@@ -0,0 +1,43 @@
...
+(def service-query-tag (env :consul-query-tag default-service-query-tag))
+
+(def env-consul-conn {:scheme      :http
+                      :server-name (env :consul-hostname default-consul-hostname)
+                      :server-port (env :consul-http-port default-consul-http-port)})
+
+(def pick-service-entry-fn rand-nth)
+
+(defn- response->address [response]
+  (let [entry (pick-service-entry-fn response)]
```
drsnyder: This is generating an index out of bounds exception:
```
[damon@dev-damon vignette]$ java -jar target/vignette-standalone.jar -p 5000 -m s3                                                                                                                                
Mode: s3. Starting server on 5000...
nil : /9ad45e83-fd30-4fad-8309-cc96295c12f6/scale-to-width/50
java.lang.IndexOutOfBoundsException
        at clojure.lang.RT.nthFrom(RT.java:795)
        at clojure.lang.RT.nth(RT.java:764)
        at clojure.core$rand_nth.invoke(core.clj:6540)
        at vignette.util.consul$response__GT_address.invoke(consul.clj:20)
        at vignette.util.consul$find_service.invoke(consul.clj:40)
        at vignette.storage.static_assets$build_url.invoke(static_assets.clj:11)
        at vignette.storage.static_assets$create_async_response_stored_object.invoke(static_assets.clj:33)
        at vignette.storage.static_assets.StaticImageStorage.get_original(static_assets.clj:44)
        at vignette.util.thumbnail$generate_thumbnail.invoke(thumbnail.clj:95)
        at vignette.util.thumbnail$get_or_generate_thumbnail.invoke(thumbnail.clj:87)
        at vignette.http.route_helpers$handle_thumbnail.invoke(route_helpers.clj:13)
        at vignette.http.routes$create_request_handlers$fn__11062.invoke(routes.clj:32)
        at compojure.core$make_route$fn__9516.invoke(core.clj:130)
        at compojure.core$wrap_route_middleware$fn__9512.invoke(core.clj:118)
        at compojure.core$if_route$fn__9488.invoke(core.clj:41)
        at compojure.core$if_method$fn__9480.invoke(core.clj:27)
        at compojure.core$routing$fn__9523.invoke(core.clj:144)
        at clojure.core$some.invoke(core.clj:2443)
        at compojure.core$routing.doInvoke(core.clj:144)
        at clojure.lang.RestFn.applyTo(RestFn.java:139)
        at clojure.core$apply.invoke(core.clj:619)
        at compojure.core$routes$fn__9527.invoke(core.clj:149)
        at compojure.core$routing$fn__9523.invoke(core.clj:144)
        at clojure.core$some.invoke(core.clj:2443)
        at compojure.core$routing.doInvoke(core.clj:144)
        at clojure.lang.RestFn.invoke(RestFn.java:423)
        at vignette.http.routes$all_routes$fn__11069.invoke(routes.clj:65)
```

src/vignette/storage/static_assets.clj
pchojnacki: S3 is proprietary protocol that isn't well standardised to begin with. I've found problems where RadosGW couldn't be used with newer s3 libraries because of encryption scheme change in the upstream S3.

I'm all for reusing technologies but we would have to spend a lot of time to reimplement S3 interface. And the only thing we would gain from this is maybe the possibility of reusing some existing libraries. 

If you guys think that us reimplementing S3 protocol is a must have feature of SAS we should talk about it more. But this is a topic for another discussion. 

Additionally the existing inherited from MW file path scheme and bucket abstraction is something SAS doesn't need to support.

src/vignette/storage/static_assets.clj
pchojnacki: Yeah its basically a GET but there is some S3 specific signing mojo included if you supply credentials. Without that there would be problems accessing non public Ceph/S3 files which is a problem guys implementing Mobile Config service encountered. Where they implemented access using simple HTTP GET.

But yeah once I finish the work required for Static Assets I plan to make some simple benchmarks for S3 code to see if the increase in throughput is worth the investment.

src/vignette/storage/static_assets.clj
ljagiello: yes, S3 interface is a must have.

pchojnacki: I think there is some misunderstanding here.
SAS is already talking via S3 to RadosGW and it definately stay that way.

What I meant was for Static Asset service to expose S3 like interface to its clients.

ljagiello: Oh that change a lot :) @pchojnacki I don't really care how services will talk with SAS. S3 interface is only case between SAS <-> CEPH.

src/vignette/setup.clj
```diff
@@ -0,0 +1,30 @@
...
+  (if (= (:mode opts) "local")
+    (do
+      (i/create-integration-env)
+      (create-local-storage-system i/integration-path))
+    (create-s3-storage-system storage-creds)))
+
+(defn create-stores [opts]
+  {
+   :wikia-store   (create-image-storage (create-object-storage opts) (:cache-thumbnails opts))
+   :static-store (sa/->StaticImageStorage)
```
drsnyder: Minor: we have been hiding the record creation behind functions in the storage class e.g. `create-static-image-storage-system`. It creates a cleaner constructor and allows you to just export that function. Please provide a similar interface for `StaticImageStorage`.

dev/user.clj
```diff
@@ -22,33 +22,35 @@
...
+                            :original     "ropes.jpg"
+                            :mode         "resize"
+                            :height       "10"
+                            :width        "10"})
 
-(def los  (create-local-storage-system itg/integration-path))
-(def lis  (create-image-storage los))
+(def los (create-local-storage-system itg/integration-path))
+(def lis (create-image-storage los))
+(defn create-stores [wikia-store] {:wikia-store wikia-store :static-store (sa/->StaticImageStorage)})
```
drsnyder: Minor: looks like this is duplicated in `setup`.

src/vignette/util/image_response.clj
```diff
@@ -38,13 +38,14 @@
...
-       (header "Content-Length" (content-length image))
-       (header "ETag" (etag image))
-       (header "X-Thumbnailer" "Vignette")
-       (add-content-disposition-header image-map)
-       (add-surrogate-header image-map)))
+
+   (as-> (response (->response-object image)) resp
+         (if-let [ct (content-type image)] (header resp "Content-Type" ct) resp)
+         (if-let [cl (content-length image)] (header resp "Content-Length" cl) resp)
+         (if-let [etag (etag image)] (header resp "ETag" etag) resp)
```
drsnyder: Minor: shouldn't all of the image maps have these headers? Also the inlined `if-let`s are hard to follow. I think I would prefer a `header` wrapper than was a noop if the input was nil so the readability could be maintained.

Something like
```clojure
(defn if-header
   [resp key val]
   (if val
       (header resp key val)
       resp))
```

src/vignette/util/image_response.clj
```diff
@@ -56,14 +57,22 @@
        (add-content-disposition-header image-map)
        (add-surrogate-header image-map)))
 
+(defn- base-filename [image-map object]
+  (or (if-let [orig (if image-map (original image-map))]
+        (string/replace orig "\"" "\\\""))
+      (if object (filename object))))
```
drsnyder: Should this last `if` be `when`? It's missing the second block.

src/vignette/util/image_response.clj
pchojnacki: Those ifs are actually from time when SAS was under active development and didn't provide some of the headers that resulted in errors.
I left them because if such situation were to ever happen again this would allow vignette to still show images.

pchojnacki:
@drsnyder @nmonterroso You are my only hope.

src/vignette/util/consul.clj
```diff
@@ -0,0 +1,49 @@
...
+  ([consul service tag]
+   (response->address (query-service consul service tag))))
+
+(defn ->uri [service] (str "http://" (service :address) ":" (service :port)))
+
+(defn build-static-asset-url [oid]
+  (str
+    (->uri
+      (find-service
+        create-consul "static-assets" service-query-tag)) "/image/" oid))
```
drsnyder: Minor: this looks like it belongs in a util ns somewhere and not in the `consul` ns.

src/vignette/util/image_response.clj
```diff
@@ -15,6 +15,11 @@
          add-surrogate-header
          surrogate-key)
 
+(defmacro when-header-val
+  ([resp key val]
+   `(if ~val
+     (header ~resp ~key ~val) ~resp)))
+
```
drsnyder: [Don't write a macro if a function will do it.](https://github.com/bbatsov/clojure-style-guide#macros).

README.md
```diff
@@ -65,6 +66,9 @@ Below is a list of environment variables that will affect the vignette runtime.
  * `ENABLE_ACCESS_LOG`            enable the NCSA access log [false]
  * `ACCESS_LOG_FILE`              NCSA acces log file [/tmp/Vignette-access.log]
  * `IMAGEMAGICK_BASE`             path to the root of the ImageMagick installation [/usr/local]
+ * `CONSUL_HOSTNAME`              Consul Agent's address [localhost]
+ * `CONSUL_HTTP_PORT`             Consul Agent's port [8500]
+ * `CONSUL_QUERY_TAG`             Default tag that will be used to query instances [production]
```
nmonterroso: We're launching services with `prod` now, has that been done for the static asset service?

src/vignette/util/image_response.clj
```diff
@@ -38,32 +43,39 @@
...
 (defn add-content-disposition-header
-  [response-map image-map]
-  (if (original image-map)
-    (let [requested-format (query-opt image-map :format)
-          filename (cond-> (string/replace (original image-map) "\"" "\\\"")
-                     requested-format (str "." requested-format))]
-      (header response-map "Content-Disposition" (format "inline; filename=\"%s\"" filename)))
-    response-map))
+  ([response-map image-map object]
```
drsnyder: What is the object parameter? It's not intuitive to me why the additional object is required.

src/vignette/util/image_response.clj
```diff
@@ -38,32 +43,39 @@
...
-       (header "X-Thumbnailer" "Vignette")
-       (add-content-disposition-header image-map)
-       (add-surrogate-header image-map)))
+   (as-> (response (->response-object image)) resp
+         (when-header-val resp "Content-Type" (content-type image))
+         (when-header-val resp "Content-Length" (content-length image))
+         (when-header-val resp "ETag" (etag image))
+         (header resp "X-Thumbnailer" "Vignette")
+         (add-content-disposition-header resp image-map image)
+         (add-surrogate-header resp image-map)))
```
drsnyder: I still don't understand why this block should change. All responses should have these headers. An image (or css, or any other static asset) should always have these.

src/vignette/util/image_response.clj
```diff
@@ -38,32 +43,39 @@
...
 
 (defn create-head-response
   [image-map]
   (-> (response "")
-       ;(header "Content-Type" (content-type image))
        (header "X-Thumbnailer" "Vignette")
        (add-content-disposition-header image-map)
        (add-surrogate-header image-map)))
 
+(defn- base-filename [image-map object]
```
drsnyder: What is the object in this case? Shouldn't we always have an `image-map`?

src/vignette/http/api_routes.clj
```diff
@@ -0,0 +1,48 @@
...
+                   :wikia wikia-regex
+                   :image-type image-type-regex
+                   :top-dir top-dir-regex
+                   :middle-dir middle-dir-regex])
+
+(def uuid-context ["/:uuid"
+                   :uuid uuid-regex])
+
+(defmacro def-api-context [ctx store]
+  `(context ~(eval ctx) [] (apply routes (api-routes ~store))))
```
drsnyder: Can this also be a function instead of a macro?

src/vignette/util/image_response.clj
```diff
@@ -38,32 +43,39 @@
...
 (defn create-image-response
   ([image image-map]
-   (-> (response (->response-object image))
-       (header "Content-Type" (content-type image))
-       (header "Content-Length" (content-length image))
-       (header "ETag" (etag image))
-       (header "X-Thumbnailer" "Vignette")
-       (add-content-disposition-header image-map)
-       (add-surrogate-header image-map)))
+   (as-> (response (->response-object image)) resp
```
nmonterroso: why change this to use `as->`? from the changes looks like `->` still works

src/vignette/http/api_routes.clj
```diff
@@ -0,0 +1,48 @@
...
+  (flatten
+    [(create-request-handlers store proto/scale-to-width-route handle-thumbnail route->thumbnail-auto-height-map)
+     (create-request-handlers store proto/scale-to-width-down-route handle-thumbnail route->thumbnail-auto-height-map)
+     (create-request-handlers store proto/scale-to-height-down-route handle-thumbnail route->thumbnail-auto-width-map)
+     (create-request-handlers store proto/window-crop-route handle-thumbnail route->thumbnail-auto-height-map)
+     (create-request-handlers store proto/window-crop-fixed-route handle-thumbnail route->thumbnail-map)
+     (create-request-handlers store proto/thumbnail-route handle-thumbnail route->thumbnail-map)
+     (create-request-handlers store proto/original-route handle-original route->original-map)]))
+
+(def wiki-context ["/:wikia:image-type/:top-dir/:middle-dir/:original/revision/:revision"
```
nmonterroso: is something either in `wiki-context` or `uuid-context`? what if it's neither (like the legacy thumbnail urls)?

src/vignette/util/consul.clj
```diff
@@ -0,0 +1,49 @@
+(set! *warn-on-reflection* true)
```
nmonterroso: why is this needed? does it change it for the running service as a whole, or just for this namespace?

pchojnacki: ups my bad. This is leftover from early experiments with clojure.

On Thu, Aug 27, 2015 at 8:12 PM, Nelson Monterroso <notifications@github.com

src/vignette/http/api_routes.clj
pchojnacki: In src/vignette/setup.clj:25
```(defn image-routes [stores]

  (list
    (def-api-context wiki-context (:wikia-store stores))
    (def-api-context uuid-context (:static-store stores))
    (hlr/legacy-routes (:wikia-store stores))))```

configures the routes for legacy URLs

On Thu, Aug 27, 2015 at 8:08 PM, Nelson Monterroso <notifications@github.com
> wrote:

> In src/vignette/http/api_routes.clj
> <https://github.com/Wikia/vignette/pull/95#discussion_r38126864>:
>
> > +                  (route->map-fn (:route-params request)
> > +                                 request)
> > +                  request))])
> > +
> > +(defn api-routes
> > +  [store]
> > +  (flatten
> > +    [(create-request-handlers store proto/scale-to-width-route handle-thumbnail route->thumbnail-auto-height-map)
> > +     (create-request-handlers store proto/scale-to-width-down-route handle-thumbnail route->thumbnail-auto-height-map)
> > +     (create-request-handlers store proto/scale-to-height-down-route handle-thumbnail route->thumbnail-auto-width-map)
> > +     (create-request-handlers store proto/window-crop-route handle-thumbnail route->thumbnail-auto-height-map)
> > +     (create-request-handlers store proto/window-crop-fixed-route handle-thumbnail route->thumbnail-map)
> > +     (create-request-handlers store proto/thumbnail-route handle-thumbnail route->thumbnail-map)
> > +     (create-request-handlers store proto/original-route handle-original route->original-map)]))
> > +
> > +(def wiki-context ["/:wikia:image-type/:top-dir/:middle-dir/:original/revision/:revision"
>
> is something either in wiki-context or uuid-context? what if it's neither
> (like the legacy thumbnail urls)?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Wikia/vignette/pull/95/files#r38126864>.
>

src/vignette/http/api_routes.clj
pchojnacki: Yes its necessary to overcome context macro limitations.

src/vignette/util/image_response.clj
pchojnacki: we have image-map but it only contains relevant filename data for wikia old requests. In SAS requests there is no filename and thus it has to be taken from elsewhere.

drsnyder: What is the filename in the case of the SAS request? Can it be derived from the image map?

If not it seems like this is something that should be handled with polymorphism. That is we should create a protocol and add the `filename` to the `image-map`.

pchojnacki: filename is taken from the stored object protocol if its not provided through image-map.
While I agree that current wikia image-map needs to be rethought I don't necessarily think that polymorphism will be the solution especially to this particualr case. 

Right now if image-map provides the filename its used here if not then stored object is used as sort of fallback.

I thought about removing filename from image map and rely only on storage object but I would like to make limit the scope of single PR of such core component that is Vignette.

drsnyder:
@pchojnacki ok, LGTM. I think we need to plan for some refactoring after you merge. The two things I think we need to address are:
 * introduction of http-kit
 * the `filename` inconsistency coming from both the `image-map` and the `object`

PR #96, opened 2015-08-31
Fix route concatenation in image routes

pchojnacki:
After deploying to prod some healthchecks started failing as well as 404's were not handled correctly.
This PR fixes that issue

@macbre


michalroszka:
:+1:

PR #97, opened 2015-08-31
Second attempt at fixing Routing concatenation

pchojnacki:
```hlr/legacy-routes``` return list of routes which in turn caused all following routes to be overriden.
This change fixes concatenation of routes list so instead of list like:
```[ route_a route_b [sub_a sub _b] ]```
We get ```[route_a route_b sub_a sub_b]```

Previous change was doing the same but it only fixed tests :disappointed:. Here I also make tests use vignette.setup as a source of sample routing configuration which aligns it with how production is configured.
@macbre @michalroszka

### DIFF
--- a/src/vignette/setup.clj
+++ b/src/vignette/setup.clj
@@ -23,7 +23,8 @@
    })
 
 (defn image-routes [stores]
-  (list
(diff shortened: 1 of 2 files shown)


macbre:
:+1:

PR #98, opened 2015-09-01
OPS-7063 increase timeouts and add documentation about them

pchojnacki:
A lot of timouts are happening especially around the time when thumbnails are GC'ed this attempts to alleviate this issue

/cc: @adamw-wikia @michalroszka @macbre @wladekb


macbre:
LGTM

PR #99, opened 2015-10-15
PLATFORM-1541 Add original filename suffix to temporary file && fix Mime detection

pchojnacki:
Without proper filename suffix both Tika/Pantomime and Imagick mime detection failed to work properly especially for SVG images.

This change reintroduces filename suffix in temp file and allows for both detection to correctly handle SVG files.

PR #101, opened 2015-10-22
PLATFORM-1541 Fix vignette SVG image scaling

pchojnacki:
this PR includes fixes for bug introduced in #99 
Addtionally I added better handling of content disposition headers that might be utf-8 encoded

### DIFF
--- a/src/vignette/storage/s3.clj
+++ b/src/vignette/storage/s3.clj
@@ -91,8 +91,9 @@
     (when-let [object (safe-get-object (add-timeouts :get (:creds this)) bucket path)]
       (when (valid-s3-get? object)
         (let [stream (:content object)
-              meta-data (:metadata object)]
-          (create-stored-object stream meta-data)))))
+              meta-data (:metadata object)
+              file-name (last (clojure.string/split (:key object) #"/"))]
+          (create-stored-object stream meta-data file-name)))))
   (put-object [this resource bucket path]
     (let [file (file-stream resource)
           mime-type (content-type resource)]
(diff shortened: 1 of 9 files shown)


src/vignette/storage/s3.clj
```diff
@@ -91,8 +91,9 @@
     (when-let [object (safe-get-object (add-timeouts :get (:creds this)) bucket path)]
       (when (valid-s3-get? object)
         (let [stream (:content object)
-              meta-data (:metadata object)]
-          (create-stored-object stream meta-data)))))
+              meta-data (:metadata object)
+              file-name (last (clojure.string/split (:key object) #"/"))]
```
drsnyder: Could this be `nil`? What should the behavior be in that case? Should we use the hash?

drsnyder:
How does this address the bug introduced in #99?

src/vignette/storage/s3.clj
pchojnacki: Good point but since :key object is the id of the object then if object is found it should have :key property set.

pchojnacki:
#99 worked by adding original filename to UUID for temp filename but that caused problems with string escaping in bash for some cases. Since it is sufficient to only use extension for the imagick/pantomime mime type detection to work correctly.  Here I've change that code to only append extension to the uuid.

src/vignette/util/thumbnail.clj
```diff
@@ -109,10 +110,13 @@
...
             "unable to get original for thumbnailing")))
 
+(defn- extract-extension [filename]
+  (last (split filename #"\.")))
+
 (defn original->local
   "Take the original and make it local."
   [original]
-  (let [temp-file (io/file (temp-filename))]
+  (let [temp-file (io/file (temp-filename nil (extract-extension (filename original))))]
```
drsnyder: Ah, ok, this looks like the part that fixes the previous PR. Can you add tests for this behavior?

drsnyder:
Minor comment, otw, LGTM.

PR #104, opened 2015-11-05
PLATFORM-1643 Allow handling of avatars without any name in metadata

pchojnacki:
Some badly uploaded avatars currently have empty file names.
This allows to handle those edge cases.

### DIFF
--- a/src/vignette/storage/static_assets.clj
+++ b/src/vignette/storage/static_assets.clj
@@ -7,7 +7,7 @@
 (defn- parse-content-disp
   [header]
   (when-let [m (if header (re-find #"(?i)filename=\"(.*)\"" header))]
-    (second m)))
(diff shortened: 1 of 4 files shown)


src/vignette/storage/static_assets.clj
```diff
@@ -7,7 +7,7 @@
 (defn- parse-content-disp
   [header]
   (when-let [m (if header (re-find #"(?i)filename=\"(.*)\"" header))]
-    (second m)))
+    (or (second m) "")))
```
jcellary: My nawias knowledge is very limited, but I'm assuming that will still create this field with "" Not sure if that's on purpose or we should just skip this field.

jcellary:
Looks good. Very pretty nawiasy

PR #105, opened 2015-11-27
PLATFORM-1584: Timeout for hanging convert

ljagiello:
Due to some bug (either ImageMagick or broken image) convert process  sometimes never finish and consuming 100% of CPU. Timeout should help for that and kill process after 5 min.

/cc @drsnyder @nmonterroso @pchojnacki

### DIFF
--- a/bin/thumbnail
+++ b/bin/thumbnail
@@ -6,6 +6,11 @@ IMAGEMAGICK_BASE=${IMAGEMAGICK_BASE:-"/usr/local"}
 CONVERT=$IMAGEMAGICK_BASE/bin/convert
 IDENTIFY=$IMAGEMAGICK_BASE/bin/identify
 
+# timeout - run a command with a time limit
+# After 5 min of converting we can assume we will never get result. Time is in seconds.
+
+TIMEOUT=${IMAGEMAGICK_BASE:-"/usr/bin/timeout"}
+TIMEOUT_TIME=${TIMEOUT_TIME:-" 300"}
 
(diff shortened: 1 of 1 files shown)


bin/thumbnail
```diff
@@ -6,6 +6,10 @@ IMAGEMAGICK_BASE=${IMAGEMAGICK_BASE:-"/usr/local"}
 CONVERT=$IMAGEMAGICK_BASE/bin/convert
 IDENTIFY=$IMAGEMAGICK_BASE/bin/identify
 
+# timeout - run a command with a time limit
+# After 5 min of converting we can assume we will never get result. Time is in seconds.
+TIMEOUT=${IMAGEMAGICK_BASE:-"/usr/bin/timeout"}
```
pchojnacki: typo here should be TIMEOUT instead of IMAGEMAGICK_BASE

pchojnacki:
LGTM

bin/thumbnail
ljagiello: Good catch! Fixed.

drsnyder:
This should be handled [here](https://github.com/Wikia/vignette/blob/master/bin/thumbnail#L33) with the `-limit time 60` directive. We aren't overriding `CONVERT_CONSTRAINTS`. In theory `convert` should not be running longer than 60s.

ljagiello:
@drsnyder `-limit time 60` is not working, there are multiple examples in the ticket.

PR #106, opened 2015-12-01
Fix detecting animated GIFs after output format of identify has changed

wladekb:
`identify` version 6.9.0-4 started to output a separate entry per each animation frame instead of a single value when run using the following command line:
`identify -format %n ....`

The change solves the issue and handles both old and new output correctly.

/cc @pchojnacki

### DIFF
--- a/bin/thumbnail
+++ b/bin/thumbnail
@@ -404,7 +404,7 @@ function animated_thumb_params() {
 
 
 function is_animated() {
-	num=$($IDENTIFY -format "%n" "$1")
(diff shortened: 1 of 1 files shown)


pchojnacki:
LGTM

ljagiello:
@wladekb is that mean we can go back to the latest ImageMagick 6.9.2?

wladekb:
@ljagiello It means we could try. Are you brave enough? :>

ljagiello:
YOLO!

PR #107, opened 2016-01-13
PLATFORM-1331 Make vignette ask Image review service for static-assets images status

pchojnacki:
To allow more flexibility in review process we decided to implement filtering of rejected static-asset images in vignette.
This change makes static-assets handling code to at the same time ask two services then resolve both responses  asynchronously and check if both services return proper responses.

Additionally, static assets images were missing SurrogateKey which this change also adds.

/cc:
@drsnyder @nmonterroso

### DIFF
--- a/src/vignette/storage/static_assets.clj
+++ b/src/vignette/storage/static_assets.clj
@@ -24,7 +24,7 @@
 
 (defn create-async-response-stored-object [static-image-get image-review-get]
   (let [static-image-response @static-image-get]
-    (if (and (-> static-image-response :status (= 200)) (-> @image-review-get :status (= 200)) )
+    (if (and (-> static-image-response :status (= 200)) (-> @image-review-get :status (= 200)))
       (->AsyncResponseStoredObject static-image-response))))
 
 (defrecord StaticImageStorage [static-image-url image-review-url] ImageStorageProtocol
@@ -40,7 +40,9 @@
 
   (original-exists? [_ image-map] nil
     (if-let [uuid (:uuid image-map)]
-      (let [response @(http/head (static-image-url uuid))] (-> response :status (= 200))))))
+      (let [static-image-response (http/head (static-image-url uuid)) image-review-response (http/get (image-review-url uuid))]
+        (and (-> @static-image-response :status (= 200))
+             (-> @image-review-response :status (= 200)))))))
 
 (defn create-static-image-storage [static-image-url image-review-url]
   (->StaticImageStorage static-image-url image-review-url))
--- a/src/vignette/storage/static_assets.clj
+++ b/src/vignette/storage/static_assets.clj
@@ -22,23 +22,25 @@
     (fs/file-exists? to))
   (->response-object [this] (file-stream this)))
 
-(defn create-async-response-stored-object [url]
-  (let [response @(http/get url {:as :stream})]
+(defn create-async-response-stored-object [static-image-get image-review-get]
+  (let [response @static-image-get]
     (if (-> response :status (= 200))
       (->AsyncResponseStoredObject response))))
 
-(defrecord StaticImageStorage [create-url] ImageStorageProtocol
+(defrecord StaticImageStorage [static-image-url image-review-url] ImageStorageProtocol
   (save-thumbnail [_ _ _] nil)
   (get-thumbnail [_ _] nil)
   (save-original [_ _ _] nil)
 
   (get-original [_ original-map]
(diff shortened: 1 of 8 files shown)


drsnyder:
Is there a jira ticket associated with this PR?

src/vignette/storage/static_assets.clj
```diff
@@ -22,23 +22,27 @@
     (fs/file-exists? to))
   (->response-object [this] (file-stream this)))
 
-(defn create-async-response-stored-object [url]
-  (let [response @(http/get url {:as :stream})]
-    (if (-> response :status (= 200))
-      (->AsyncResponseStoredObject response))))
+(defn create-async-response-stored-object [static-image-get image-review-get]
+  (let [static-image-response @static-image-get]
```
drsnyder: Why do both have to respond with a 200 to return the `static-image-response`?

src/vignette/storage/static_assets.clj
```diff
@@ -22,23 +22,27 @@
...
     (if-let [uuid (:uuid original-map)]
-      (create-async-response-stored-object (create-url uuid))))
+      (create-async-response-stored-object
+        (http/get (static-image-url uuid) {:as :stream})
+        (http/get (image-review-url uuid)))))
 
   (original-exists? [_ image-map] nil
     (if-let [uuid (:uuid image-map)]
-      (let [response @(http/head (create-url uuid))] (-> response :status (= 200))))))
+      (let [static-image-response (http/head (static-image-url uuid)) image-review-response (http/get (image-review-url uuid))]
```
drsnyder: Why is one a `HEAD` and the other a `GET`?

src/vignette/util/image_response.clj
```diff
@@ -92,7 +94,9 @@
...
   [image-map]
-  (try
-    (digest/sha1 (fully-qualified-original-path image-map))
-    (catch Exception e
-      (str "vignette-"(:original image-map)))))
+  (if-let [sk (:uuid image-map)] sk
+    (try
+      (digest/sha1 (fully-qualified-original-path image-map))
+      (catch Exception e
+        (print e)
```
drsnyder: Did you mean to leave the `print` here? Do you want to log it instead?

src/vignette/storage/static_assets.clj
```diff
@@ -22,23 +22,27 @@
...
   (->response-object [this] (file-stream this)))
 
-(defn create-async-response-stored-object [url]
-  (let [response @(http/get url {:as :stream})]
-    (if (-> response :status (= 200))
-      (->AsyncResponseStoredObject response))))
+(defn create-async-response-stored-object [static-image-get image-review-get]
+  (let [static-image-response @static-image-get]
+    (if (and (-> static-image-response :status (= 200)) (-> @image-review-get :status (= 200)))
+      (->AsyncResponseStoredObject static-image-response))))
```
drsnyder: Minor: it would be nice to have a method that will either dereference the future or pass through a map so that you don't have to wrap the response in a `future` in cases where it's not needed e.g. the tests. Something like:

```clojure
(def deref-request [f]
   (if (future? f)
       (deref f)
       f))
```

src/vignette/storage/static_assets.clj
pchojnacki: Call to image review ensures that image exists and has any of the specified flags, i.e. it checks if the image is safe to show.
So to show the image both services need to agree that image exists and can be shown.

src/vignette/storage/static_assets.clj
pchojnacki: I prefer to specifically deref variables if possible to know which variables are futures and which are not

pchojnacki:
ping @drsnyder @nmonterroso

drsnyder:
On the surface, it doesn't seem like this logic belongs in vignette. It's not related to thumbnailing. Are we sure this is where we want to put it?

pchojnacki:
If you consider vignette as end user image server, sort of front end.
Its not uncommon in  SOA to do those aggregations of calls to different services on 'frontend' servers. Checking if the content is reviewed and can be shown fits quite nicely with that flow. 
I guess we could decouple that to a separate service that does reviewed-static-image-serving. But I'd like to avoid that as it would be rather overcomplicating things. 
Anyway I'm open to suggestion and since you're still in PL tomorrow we should chat about it.

pchojnacki:
@drsnyder Do you think its ok to merge it now ?

drsnyder:
LGTM

PR #108, opened 2016-01-14
PLATFORM-1805: Fix top-crop-down when requested size is smaller than original image

wladekb:
Theoretically behavior for "top-crop-down" and "zoom-crop-down" was unspecified when requested dimensions were smaller than those of the original image. Practically the output was a cropped image but the image dimensions ratio was changed in a non-intuitive (and non-practical) way.

This change brings a bit more of sanity into this use case. The highest priority in the current implementation is to preserve the ratio of the requested dimensions.

https://wikia-inc.atlassian.net/browse/PLATFORM-1805

/cc @drsnyder @nmonterroso @SzymonCierniewski @pchojnacki

### DIFF
--- a/bin/thumbnail
+++ b/bin/thumbnail
@@ -298,28 +298,35 @@ function recalculate_requested_size_for_crop_down() {
 	# need to use bc for floating-point arithmetic below
 	local CODE=`cat <<EOF
 scale = 20
-define i(x) {
-    auto s
-    s = scale
+define trunc(x) {
+    auto saved_scale
+    saved_scale = scale
     scale = 0
     x /= 1   /* round x down */
-    scale = s
+    scale = saved_scale
     return (x)
 }
 
-s = 1
-u = $origwidth / $WIDTH
-if (u<s) s = u
+define round(x) {
+	x += 0.5 /* round */
+	x += 0.000000001 /* epsilon */
+	return trunc(x)
+}
 
-u = $origheight / $HEIGHT
-if (u<s) s = u
+min_acceptable_scale = 1
+candidate_scale = $origwidth / $WIDTH
+if (candidate_scale < min_acceptable_scale) min_acceptable_scale = candidate_scale
 
-i($WIDTH * s + 0.000000001)
-i($HEIGHT * s + 0.000000001)
+candidate_scale = $origheight / $HEIGHT
+if (candidate_scale < min_acceptable_scale) min_acceptable_scale = candidate_scale
 
-s
-$WIDTH * s
-$HEIGHT * s
+round($WIDTH * min_acceptable_scale)
+round($HEIGHT * min_acceptable_scale)
+
+/* debug information - ignored on production */
+min_acceptable_scale
+$WIDTH * min_acceptable_scale
+$HEIGHT * min_acceptable_scale
 EOF
 `
 	local CALCOUT=`echo "$CODE" | bc`
--- a/bin/thumbnail
+++ b/bin/thumbnail
@@ -285,6 +285,49 @@ function set_custom_type() {
 	OUT="${1}:${OUT}"
 }
 
+function recalculate_requested_size_for_crop_down() {
+	# recalculate output image dimensions in cases of "...-crop-down" modes
+	# in such a way that an output image has the same aspect ratio as request
+	# after executing imagemagick
+	local origwidth=$1
+	local origheight=$2
+	local CODE
+	local x
+	local y
+
+	# need to use bc for floating-point arithmetic below
+	local CODE=`cat <<EOF
+scale = 20
+define i(x) {
+    auto s
+    s = scale
+    scale = 0
+    x /= 1   /* round x down */
+    scale = s
+    return (x)
+}
+
+s = 1
+u = $origwidth / $WIDTH
+if (u<s) s = u
+
+u = $origheight / $HEIGHT
+if (u<s) s = u
+
+i($WIDTH * s + 0.000000001)
+i($HEIGHT * s + 0.000000001)
+
+s
+$WIDTH * s
+$HEIGHT * s
+EOF
+`
+	local CALCOUT=`echo "$CODE" | bc`
+
+	WIDTH=`echo "$CALCOUT" | head -n 1`
+	HEIGHT=`echo "$CALCOUT" | head -n 2 | tail -n 1`
+}
+
 function thumb_params() {
 	QUALITY=''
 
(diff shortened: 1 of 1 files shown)


bin/thumbnail
```diff
@@ -285,6 +285,49 @@ function set_custom_type() {
...
+
+s = 1
+u = $origwidth / $WIDTH
+if (u<s) s = u
+
+u = $origheight / $HEIGHT
+if (u<s) s = u
+
+i($WIDTH * s + 0.000000001)
+i($HEIGHT * s + 0.000000001)
```
drsnyder: Why do you have to add the float?

bin/thumbnail
```diff
@@ -285,6 +285,49 @@ function set_custom_type() {
...
+define i(x) {
+    auto s
+    s = scale
+    scale = 0
+    x /= 1   /* round x down */
+    scale = s
+    return (x)
+}
+
+s = 1
```
nmonterroso: Can we change these variable names to be a bit more descriptive?

bin/thumbnail
```diff
@@ -285,6 +285,49 @@ function set_custom_type() {
...
+u = $origwidth / $WIDTH
+if (u<s) s = u
```
drsnyder: Please use more descriptive names for `s` (`aspectRatio`?) and `u` (`widthRatio`/`heightRatio`?).

bin/thumbnail
```diff
@@ -391,10 +433,9 @@ function thumb_params() {
 			echo "Height ${origheight} x Width ${origwidth}"
 			if [[ $origheight -gt $HEIGHT && $origwidth -gt $WIDTH ]]; then
 				THUMB_PARAMS="${QUALITY} -thumbnail ${WIDTH}x${HEIGHT}+0+0^ -gravity center -crop ${WIDTH}x${HEIGHT}+0+0!"
-			elif [ $origheight -gt $HEIGHT ] || [ $origwidth -gt $WIDTH ]; then
-				THUMB_PARAMS="${QUALITY} -gravity center -crop ${WIDTH}x${HEIGHT}+0+0! -thumbnail ${WIDTH}x${HEIGHT}+0+0"
 			else
-				THUMB_PARAMS="${QUALITY}"
+				recalculate_requested_size_for_crop_down $origwidth $origheight
+				THUMB_PARAMS="${QUALITY} -gravity center -crop ${WIDTH}x${HEIGHT}+0+0 -thumbnail ${WIDTH}x${HEIGHT}+0+0"
```
drsnyder: The imagemagick documentation isn't always clear but I wonder if the above is the same as
```
THUMB_PARAMS="${QUALITY} -gravity center -crop ${WIDTH}x${HEIGHT}+0+0> -thumbnail ${WIDTH}x${HEIGHT}+0+0"
```

I only replaced the `!` with `>`. From [documentation](http://www.imagemagick.org/script/command-line-processing.php#geometry) the `>` will only shrink an image with dimension(s) larger than the width and height provided.

It would be nice to be able to lean on imagemagick as much as possible to simplify the logic in the bash script.

jcellary:
Yey, @wladekb go for clojure :stuck_out_tongue_closed_eyes:

bin/thumbnail
wladekb: This is required due to rounding issues that are common in floating-point operations. Moreover there were a lot of cases where the value of ``$WIDTH * s`` resulted in eg. ``599.99999999999235825`` which should become ``600``. The value of ``0.000000001`` is often described as ``epsilon`` which you use just before truncating the fraction point so you get results closer to what you wanted.

bin/thumbnail
wladekb: When I looked in the documentation of ``bc`` I found that variables can only be one single lower letter but I'll double check if that's the case in the version we have on our servers.

bin/thumbnail
wladekb: See note above.

bin/thumbnail
wladekb: It seems I misunderstood the documentation. Fixed.

bin/thumbnail
wladekb: The main reason for undesired behavior of thumbnail script was not keeping an aspect ratio. While scaling is well-supported in imagemagick I couldn't find a way to request a crop operation that would give us what we wanted here *without* upscaling the image which is not a desired side-effect in a "down" mode. The main issue I don't know how to solve is that your highest priority is keeping the aspect ratio of the generated image in line with the requested aspect ratio and I couldn't find any working way to specify the aspect ratio in imagemagick's `geometry`.

http://www.imagemagick.org/script/command-line-processing.php#geometry

I've spent some time on finding a working solution but failed and fell back to do a workaround using ``bc`` @drsnyder  Do you want to give it a try?

wladekb:
@nmonterroso @drsnyder I changed the `bc` script to improve readability. How does it look like now?

bin/thumbnail
drsnyder: Ok, I tried this morning and could not get the behavior expected by @SzymonCierniewski out of imagemagic.

bin/thumbnail
drsnyder: That seems dicey. The function you have above to round down could probably be adapted for this purpose. See also this [SO](http://stackoverflow.com/questions/26861118/rounding-numbers-with-bc-in-bash) post.

drsnyder:
`top-crop-down` was originally an attempt to best-effort support for the top cropping provided by the old thumbnailer. Historically, both the usage and the effectiveness where limited. The placements where I saw this feature being used rarely looked good. At best it's a coarse approximation to face cropping.

To be honest, I'm a little torn about seeing the use of this feature expanded and time and effort put into improving and maintaining it. Do we _really_ need it?

drsnyder:
Other than the comment regarding rounding, LGTM. Thanks for doing this @wladekb.

bin/thumbnail
wladekb: Please see the updated version where I moved this into the new "round" function which is more readable and intuitive.

wladekb:
I'm pretty sure we understood the "top-crop-down" this way:
"Give me an image fragment that I can put into an area of WIDTH x HEIGHT. I don't want any empty places in the image and if you need to cut part of the image make sure to include the top part of the image."

And that also required the desired width and height be smaller than the ones of original image. I'm not trying to argue with the selection of the part of image because it's a separate issue and I'm now focusing on delivering the most reasonable output for this request.

When I think about the most reasonable output if the size criteria is not met i would for upscaling the image before sending it back and that's what happens in "top-crop" mode but we're talking about a "...-down" mode which explicitly forbids upscaling. However that's probably the best content I can deliver so I need to generate the same image content but I need to adjust the size of the output image and that's exactly what the current implementation does.

This requires adding a single sentence to the request description:
" In a rare case I'm able to upscale it on client side."
which is also a good choice from performance perspective as we need less bandwidth to basically deliver the same user experience for the client.

drsnyder:
@wladekb after wrestling with it some I think this is better than the current behavior. As described in the documentation I think the current behavior is a bug because it will create the thumbnail instead of returning the original.

And, I think this fix is _better_ than returning the original because it at least tries to give you something that preserves the aspect ratio. :ship:

PR #109, opened 2016-01-26
PLATFORM-1843 make imagick argument handling stop using evals

pchojnacki:
this should fix problems with special  chars in filenames as well as mitigate potential attack vector

### DIFF
--- a/bin/thumbnail
+++ b/bin/thumbnail
@@ -2,7 +2,7 @@
 # Wrapper for imagemagick that encapsulates our thumbnail logic.
 # @author Damon Snyder (damon@huddler-inc.com) 
 
-IMAGEMAGICK_BASE=${IMAGEMAGICK_BASE:-"/usr/local"}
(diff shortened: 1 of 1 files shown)


pchojnacki:
@drsnyder and @nmonterroso could you guys take a look at this Bash only PR ? pretty please 

btw travis fails seems to be unrelated so please ignore it for now.

nmonterroso:
LGTM - I don't think any tests actually touch the shell script.

drsnyder:
Most/all of the inputs to the thumbnail have to go through the routes which have to satisfy restrictive [regexes](https://github.com/Wikia/vignette/blob/master/src/vignette/util/regex.clj). Unless I'm mistaken that should serve as an effective sanitization layer. The two gaps would be the few query params and the file name. We should probably scrutinize the former. The latter has to exist on disk (and in persistent storage) so should already be escaped.

PR #112, opened 2016-02-02
PLATFORM-1808 Fix file extension handling 

pchojnacki:
some junk filenames caused junk file extension to be used on filesystem.
This PR fixes it by only allowing ([a-z_A-Z0-9]*) in file extension otherwise the extension will be treated as nil.


drsnyder:
lgtm

PR #113, opened 2016-03-13
PLATFORM-1891 fix broken thumbnails when filename is nil

pchojnacki:


macbre:
A unit test? ;)

macbre:
LGTM (because "what can possibly go wrong if we have unit tests")

pchojnacki:
LGTM

PR #114, opened 2016-04-18
Ser 81 thumb verifier

pwojcik86:
[Ticket](https://wikia-inc.atlassian.net/browse/SER-81)

This PR introduces a check of the generated thumbnail's size, so that it's possible to investigate potential errors (if any).

//cc @pchojnacki

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -3,6 +3,8 @@
   :url "http://example.com/FIXME"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
+  :repositories [["snapshots" "https://artifactory.wikia-inc.com/artifactory/libs-snapshot-local/"]
+                 ["releases" "https://artifactory.wikia-inc.com/artifactory/libs-release-local/"]]
   :dependencies [[org.clojure/clojure "1.5.1"]
                  [org.clojure/tools.cli "0.3.1"]
                  [cheshire "5.3.1"]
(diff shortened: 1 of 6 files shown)


project.clj
```diff
@@ -17,9 +19,8 @@
                  [slingshot "0.10.3"]
                  [ring/ring-devel "1.4.0"]
                  [useful "0.8.8"]
-                 [wikia/commons "0.1.3-SNAPSHOT"]
-                 [http-kit "2.1.18"]
-                 [org.slf4j/slf4j-log4j12 "1.7.12"]]
+                 [wikia/commons "0.1.5-SNAPSHOT"]
```
pchojnacki: It would be best not to use snapshot version

src/vignette/util/thumb_verifier.clj
```diff
@@ -0,0 +1,169 @@
...
+      :thumb-map thumb-map
+      }))))
+
+(def identify-bin (str (env :imagemagick-base "/usr/local/bin") "/identify"))
+
+(defn identify
+  [file]
+  (:out (sh identify-bin file)))
+
+(defn size-of
```
pchojnacki: I think you can try using Apache Tika instead of shelling out to imagick-identify

pchojnacki:
Two minor comments otherwise nice PR.

pchojnacki:
Once the SNAPSHOT is substituted for release version the it LGTM

PR #116, opened 2016-04-21
SER-81 thumb verifier rounding error fix

pwojcik86:
[Ticket](https://wikia-inc.atlassian.net/browse/SER-81)

This PR fixes the rounding error issue (too many false positives in error log) and introduces the `area_ratio` field for easier log filtering.


pchojnacki:
LGTM

PR #117, opened 2016-11-16
SER-532 Adds forwarding status query params to image-review

africazulu:
###### Change Description:
Vignette can forward only status query params to image-review

### DIFF
--- a/src/vignette/storage/static_assets.clj
+++ b/src/vignette/storage/static_assets.clj
@@ -24,7 +24,7 @@
 
 (defn create-async-response-stored-object [static-image-get image-review-get]
   (let [static-image-response @static-image-get]
-    (if (and (-> static-image-response :status (= 200)) (some #(= % (:status @image-review-get)) [200 404]))
+    (if (and (-> static-image-response :status (= 200)) (some #(not (= % (:status @image-review-get))) [403]))
       (->AsyncResponseStoredObject static-image-response))))
 
 (defrecord StaticImageStorage [static-image-url image-review-url] ImageStorageProtocol
--- a/src/vignette/storage/static_assets.clj
+++ b/src/vignette/storage/static_assets.clj
@@ -36,7 +36,9 @@
     (if-let [uuid (:uuid original-map)]
       (create-async-response-stored-object
         (http/get (static-image-url uuid) {:as :stream})
-        (http/get (image-review-url uuid)))))
+        (http/get (image-review-url uuid
+                                   (get-in original-map [:options :status])
+                                   (get-in original-map [:options :fastlyBypass]))))))
 
   (original-exists? [_ image-map] nil
     (if-let [uuid (:uuid image-map)]
--- a/src/vignette/storage/static_assets.clj
+++ b/src/vignette/storage/static_assets.clj
@@ -24,7 +24,7 @@
 
 (defn create-async-response-stored-object [static-image-get image-review-get]
   (let [static-image-response @static-image-get]
-    (if (and (-> static-image-response :status (= 200)) (some #(= % (:status @image-review-get)) [200 404 400 500]))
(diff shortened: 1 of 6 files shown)


tmszdmsk:
and... what about tests?

africazulu:
Tests are added for changed/added functionality

src/vignette/storage/static_assets.clj
```diff
@@ -24,7 +24,7 @@
 
 (defn create-async-response-stored-object [static-image-get image-review-get]
   (let [static-image-response @static-image-get]
-    (if (and (-> static-image-response :status (= 200)) (some #(= % (:status @image-review-get)) [200 404]))
+    (if (and (-> static-image-response :status (= 200)) (some #(not (= % (:status @image-review-get))) [403]))
```
nmonterroso: This seems like it's fundamentally changing what this function is doing. Why did it used to check for a `200` or `404` but now it needs to check for not a `403`? Is the other check no longer required?

src/vignette/util/query_options.clj
```diff
@@ -12,6 +13,8 @@
 (def query-opts-map {:fill (create-query-opt #"^#[a-g0-9]+$|^\w+$" true true)
                      :format (create-query-opt #"^\w+$")
                      :frame (create-query-opt #"[0-9]+" true true)
+                     :status (create-query-opt #"^\w+$")
+                     :fastlyBypass (create-query-opt uuid-regex)
```
nmonterroso: What's the purpose of `fastlyBypass`? It's not clear if it's some sort of secret key or if a random string that's passed by the user would work. Also, FWIW we tried to keep vignette as agnostic about its environment as much as we could - we tried not to worry about things like fastly, since those really shouldn't be concerns of the application. I'm pretty sure that's been broken already, but why does vignette now need something from fastly?

africazulu:
We needed some changes in vignette to be able to display rejected images in some contexts. So we need to forward some query params from vignette to image-review.

src/vignette/storage/static_assets.clj
africazulu: It seems that I changed it too quickly, allowing vignette to display an image in more than required response statuses from image-review. I've changed it once more.

nmonterroso:
Thanks for the changes - LGTM

Are failing tests related? If so we should fix those before merging.

src/vignette/storage/static_assets.clj
```diff
@@ -24,7 +24,7 @@
 
 (defn create-async-response-stored-object [static-image-get image-review-get]
   (let [static-image-response @static-image-get]
-    (if (and (-> static-image-response :status (= 200)) (some #(= % (:status @image-review-get)) [200 404]))
+    (if (and (-> static-image-response :status (= 200)) (some #(= % (:status @image-review-get)) [200 404 400]))
```
sqreek: We need to concern about what happen when image-review will not work properly and it will throws 500.

With this implementation all images will be not display.

sqreek:
tests are failing because travis cannot access to artifactory
```
Could not transfer artifact wikia:commons:pom:0.1.5 from/to snapshots (https://artifactory.wikia-inc.com/artifactory/libs-snapshot-local/): Access denied to: https://artifactory.wikia-inc.com/artifactory/libs-snapshot-local/wikia/commons/0.1.5/commons-0.1.5.pom , ReasonPhrase:Forbidden.
```

src/vignette/storage/static_assets.clj
africazulu: We could add 500, do we need any other status?

africazulu:
I put back the functionality where we do not return an image from image review only in the case if it returns 403 status. This is the safest approach.  If anything would happen to image review on production then vignette would not serve any images.

africazulu:
I've also changed the endpoint where vignette asks image-review for image status as it was changed to be more RESTful.

PR #119, opened 2017-01-13
PLATFORM-2660: log request details (path, format, image size, headers))

themech:
Logging the details of the requests so we can check (based on the accept headers and user agents) what % of the request could accept webp format.

@pwojcik86

### DIFF
--- a/src/vignette/http/middleware.clj
+++ b/src/vignette/http/middleware.clj
@@ -5,6 +5,7 @@
             [ring.util.response :refer [response status charset header get-header]]
             [slingshot.slingshot :refer [try+ throw+]]
             [vignette.util.image-response :refer :all]
+            [vignette.util.query-options :refer :all]
             [wikia.common.logger :as log]
             [wikia.common.perfmonitoring.core :as perf])
   (:import [java.net InetAddress]))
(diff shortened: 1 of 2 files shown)


src/vignette/http/middleware.clj
```diff
@@ -41,6 +42,19 @@
 (declare add-cache-control-header
          hours-to-seconds)
 
+(defn log-image-request
+  [handler]
+  (fn [request]
+    (let [response (handler request)]
+      (log/warn "Image request" {:status (get response :status 0)
```
pwojcik86: Why is this a `warn` and not `info`?

src/vignette/http/middleware.clj
```diff
@@ -41,6 +42,19 @@
...
+                             :uri (:uri request)
+                             :format (get (:query-params request) "format")
+                             :accept (get (:headers request) "accept")
+                             :agent (get (:headers request) "user-agent")
+                             :size (get (:headers response) "Content-Length" "0")
```
pwojcik86: I don't think `get` is necessary here. A map is also a function in Clojure. Unless you want to be a bit more verbose, I'd remove it.

src/vignette/http/middleware.clj
```diff
@@ -41,6 +42,19 @@
 (declare add-cache-control-header
          hours-to-seconds)
 
+(defn log-image-request
+  [handler]
+  (fn [request]
```
pwojcik86: You could use deconstruction here to get fields from the request, if you want: `{params :query-params uri :uri headers :headers :as request}`. The `:as request` part will still allow you to use the whole argument.

ludwikkazmierczak:
Tested And Accepted

PR #122, opened 2017-01-30
PLATFORM-2817: Serve WebP images based on the Accept header

themech:


src/vignette/util/query_options.clj
```diff
@@ -18,6 +18,9 @@
                      :replace (create-query-opt #"^true$" false)
                      :zone (create-query-opt #"\w+")})
 
+(def webp-accept-header-name "accept")
+(def webp-accept-header-value "image/webp")
```
Grunny: Would these be better in `route_helpers` where they are used, rather than here which is mostly focused on query parameters?

themech: Good point, I forgot to move it when moving some functions around.

src/vignette/http/route_helpers.clj
```diff
@@ -35,10 +38,31 @@
                             #"^\/(.*)"
                             "$1")))
 
+(defn route->requested-format
+  "Extract information about the format being explicitely specified in the url"
+  [request-map request]
+  (assoc request-map :requested-format (:format (extract-query-opts request))))
```
pwojcik86: What's the point of duplicating information about the requested format?

src/vignette/http/route_helpers.clj
```diff
@@ -35,10 +38,31 @@
...
+  [request options]
+  (if (empty? (:format options))
+    (add-webp-format-option-if-supported request options)
+    options))
+
 (defn route->options
   "Extracts the query options and moves them to 'request-map'"
   [request-map request]
-  (assoc request-map :options (extract-query-opts request)))
+  (assoc request-map :options (autodetect-request-format request (extract-query-opts request))))
```
pwojcik86: You can leave this function as-is and use the extracted `:format` value in `route->requested-format`

src/vignette/http/route_helpers.clj
themech: Good catch, I simplified the flow

src/vignette/http/route_helpers.clj
themech: Done in https://github.com/Wikia/vignette/pull/122/commits/8d4030e208ecd0eb4869c42f925639cbfad69ca9

harnash:
@pwojcik86 we addressed all comments. Is there anything left that we should do?

ludwikkazmierczak:
Tested and accepted

PR #124, opened 2017-02-06
PLATFORM-2818: Added support for WebP detection for full-size images

harnash:


src/vignette/util/external_hotlinking.clj
```diff
@@ -5,20 +5,32 @@
...
+(def force-webp-params {:request-type :thumbnail
+                        :thumbnail-mode "type-convert"})
 
 (defn force-thumb? [request]
   (if-let [vary-string (get-in request [:headers force-header-name])]
     (.contains vary-string force-header-val)))
 
+(defn force-webp? [image-params]
+  (if-let [format (get-in image-params [:options :format])]
+    (.contains format force-webp-val)))
```
Grunny: Could be simplified to:
```clojure
(defn force-webp? [image-params]
  (= force-webp-val (get-in image-params [:options :format])))
```

harnash:
ping: @Wikia/search-team please take a look

src/vignette/media_types.clj
```diff
@@ -15,6 +15,10 @@
   [file]
   (contains? webp-mime-types (mime-type-of (or file ""))))
```
themech: As you added the `webp-supported-content-type?`, we should either call it here like `(webp-supported-content-type? (mime-type-of (or file "")))` or make change other places to use `webp-supported-content-type?` instead of `webp-supported?`

src/vignette/util/external_hotlinking.clj
```diff
@@ -1,24 +1,41 @@
 (ns vignette.util.external-hotlinking
-  (:require [vignette.protocols :refer :all]
+  (:require [vignette.media-types :refer :all]
+            [vignette.protocols :refer :all]
             [vignette.storage.protocols :refer :all]
             [vignette.util.thumbnail :as u]))
 
 (def force-header-name "x-vignette-force-thumb")
 (def force-header-val "1")
+(def force-webp-val "webp")
```
themech: If we decided to define this value, we can move it somewhere else and use it other places where we currently have strings (like https://github.com/Wikia/vignette/blob/52a6aa9ed201b7456a1971275baca6e29d4eb030/src/vignette/http/route_helpers.clj#L48).

src/vignette/media_types.clj
harnash: @themech Fixed in 6a3cd15

src/vignette/util/external_hotlinking.clj
harnash: Fixed in 6a3cd15

src/vignette/media_types.clj
harnash: Fixed in d388c845caa8e5e1712da9b21a9912a151ec923c

ludwikkazmierczak:
Tested and Accepted

PR #129, opened 2017-02-28
PLATFORM-2920: Detect original image content type base on content, not on extension

themech:
Some of the jpg&png files are stored without an extension. When generating thumbnails, use the actual content type to make a decision about serving webp.


ludwikkazmierczak:
Accepted

PR #132, opened 2017-03-17
Platform-2890 webp for legacy

harnash:
ping: @Wikia/core-platform-team

### DIFF
--- a/src/vignette/http/legacy/route_helpers.clj
+++ b/src/vignette/http/legacy/route_helpers.clj
@@ -27,15 +27,11 @@
   (if-let [vary-string (get-in request [:headers webp-accept-header-name])]
     (.contains vary-string mt/webp-mime-type)))
 
-(defn add-webp-format-option-if-supported
+(defn autodetect-request-format
   [request options]
   (if (browser-supports-webp? request)
     (merge options {:format mt/webp-format})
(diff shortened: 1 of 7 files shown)


src/vignette/http/legacy/route_helpers.clj
```diff
@@ -21,6 +23,22 @@
 (defn archive? [map]
   (= (:zone map) (str "/" archive-dir)))
 
+(defn browser-supports-webp? [request]
```
Grunny: Since these methods basically duplicate the ones in the normal `route_helpers` file, what do you think of moving them to a shared helper file and reusing them in both?

harnash: @Grunny I was thinking about that but since legacy are in theory to be removed soon™️ I wanted this code to be self contained and easily removable

themech: > removed soon

:grinning:

harnash: Alright. Fixed in 43dbf8d

harnash:
@Grunny can we push it to tests then?

ludwikkazmierczak:
Tested and Accepted

PR #133, opened 2017-03-21
PLATFORM-2935: we do not thumbnail audio/video files in Vignette

harnash:
To properly handle ogg files I had to update pantomime library which resulted in returning "wierd" mime-types for ogg. For example:

- audio/vorbis (for audio)
- video/theora (for video)

To properly pass-through this formats in Vignette I changed the way filter detects those file so right now we pass-through all mime-types that are recognized as video or audio via: `pantomime.media/video?` or `pantomime/audio?`


test/vignette/http/integration_passthrough_test.clj
```diff
@@ -0,0 +1,59 @@
+(ns vignette.http.integration-passthrough-test
```
harnash: I had to put it in separate file because:
```
Exception in thread "main" java.lang.ClassFormatError: Invalid method Code length 66318 in class file vignette/http/integration_test$eval10458, compiling:(vignette/http/integration_test.clj:22:1)
```

PR #135, opened 2017-03-22
Platform-2946: Serve GIF images as WebP

themech:
DO NOT MERGE YET

This enables gif->webp conversion in Vignette. For animated GIF images, there is an additional step in the thumbnail script - it uses gif2webp util to do the conversion.

We should also update chef repo to deploy libwebp/gif2webp version 0.6.0 on Vignette machines.


bin/thumbnail
```diff
@@ -528,7 +546,10 @@ function thumb() {
 	is_animated "$IN"
 	animated=$?
 
-	if [[ $gif -eq 1 && $animated -eq 1 ]]; then
+	target_webp ${OUT}
+    webp=$?
```
harnash: indentation broke

bin/thumbnail
```diff
@@ -551,6 +572,13 @@ function thumb() {
 
 		if eval ${magick}; then
 			echo ${magick}
+			if [[ $gif -eq 1 && $animated -eq 1 && $webp -eq 1 ]]; then
+			    WEBP_OUT=${OUT}.webp
+			    gif2webp_result="${TIMEOUT} ${TIMEOUT_TIME} ${GIF2WEBP} ${GIF2WEBP_OPTIONS} "${OUT}" -o "${WEBP_OUT}""
+			    if eval ${gif2webp_result}; then
+			        mv ${WEBP_OUT} ${OUT}
+			    fi
```
harnash: We should return proper exit code when gif2web fails

harnash: After offline talk we decided to leave this and serve original GIFs and monitor failures if there are any. Then we decide if we should fail and return error code or not.

bin/thumbnail
```diff
@@ -568,14 +568,22 @@ function thumb() {
 		echo ${magick}
 		exec ${magick}
 	else
-		magick="${TIMEOUT} ${TIMEOUT_TIME} ${CONVERT} ${THUMB_PRE_OPTIONS} "${IN}" ${CONVERT_CONSTRAINTS} "-auto-orient" ${THUMB_PARAMS} "${OUT}""
+	    if [[ $gif -eq 1 && $animated -eq 1 && $MODE -eq "type-convert" ]]; then
+	        # For animated GIFs imagemagick is likely to exhaust resources. So for full size GIFs
+	        # when we only convert them the webp format (without extracting a frame), we can skip
+	        # the thumbnailing parameters so imagemagick will just copy the file. The type
+	        # conversion is done by gif2webp anyway.
```
Grunny: Minor: whitespace is off in these five lines mixing spaces and tabs.

PR #140, opened 2017-04-14
PLATFORM-2995: Log slow thumbnailer runs

themech:
If thumbnailer script runs for longer than 5 seconds, log the thumbnailing details so we can investigate performance issues with some of the images

### DIFF
--- a/src/vignette/util/thumbnail.clj
+++ b/src/vignette/util/thumbnail.clj
@@ -56,15 +56,15 @@
 
 (defn run-thumbnailer
   ([args thumb-map]
-  (let [start-ms (System/currentTimeMillis)
-        result (perf/timing :imagemagick (apply sh args))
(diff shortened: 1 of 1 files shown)


src/vignette/util/thumbnail.clj
```diff
@@ -55,8 +55,18 @@
           thumb-map))
 
 (defn run-thumbnailer
-  [args]
-  (perf/timing :imagemagick (apply sh args)))
+  ([args thumb-map]
+  (let [start-ms (System/currentTimeMillis)
+        result (perf/timing :imagemagick (apply sh args))
+        end-ms (System/currentTimeMillis)]
+    (if (> (- end-ms start-ms) 5000)
```
pwojcik86: nitpick: You can move `(- end-ms start-ms)` into `let` as the last value

PR #145, opened 2017-11-07
SER-1338 | Implement caching methods

slayful:


src/vignette/storage/static_assets.clj
```diff
@@ -31,10 +51,23 @@
...
+        resource
+        thumb-map
+        mt/thumbnail-path))
+
+  (get-thumbnail [this thumb-map]
+       (get* (:store this)
+         thumb-map
+         mt/thumbnail-path))
+
+  (save-original [this resource original-map]
```
themech: When does vignette have to save static assets originals? I thought SA caching was needed only for thumbnails.

src/vignette/storage/static_assets.clj
```diff
@@ -31,10 +51,23 @@
...
+  (save-thumbnail [this resource thumb-map]
+      (put* (:store this)
+        resource
+        thumb-map
+        mt/thumbnail-path))
+
+  (get-thumbnail [this thumb-map]
+       (get* (:store this)
+         thumb-map
+         mt/thumbnail-path))
```
themech: `thumbnail-path` creates a path from keys such as `top-dir` and  `middle-dir`. Do we have these in `thumb-map` for static assets?

src/vignette/media_types.clj
```diff
@@ -152,6 +152,16 @@
         window-width (:window-width data)]
     (window-format width x-offset window-width)))
 
+(defn static-assets-thumb-map->path [data image-path]
+      (let [prefix (thumb-map->prefix data)
+            name (format "%spx-%spx-%s%s"  (or (width data) "X") (or (height data) "X") (mode data) (query-opts-str data))]
```
slayful: so if width or height is not present it's gonna map to `Xpx-Xpx`?

slayful:
LGTM, but don't we want to add a test for caching?

src/vignette/storage/s3.clj
```diff
@@ -105,6 +105,8 @@
   (put-object [this resource bucket path]
     (let [file (file-stream resource)
           mime-type (content-type resource)]
+      (if (perf/timing :s3-get (not (s3/bucket-exists? creds bucket)))
```
pwojcik86: I wouldn't time bucket read/creation in the same measurement as file reads/writes

src/vignette/storage/static_assets.clj
```diff
@@ -1,8 +1,33 @@
...
   (:require [vignette.storage.protocols :refer :all]
+            [slingshot.slingshot :refer [throw+]]
             [org.httpkit.client :as http]
             [vignette.util.filesystem :as fs]
-            [clojure.java.io :as io]))
+            [clojure.java.io :as io]
+            [vignette.media-types :as mt]))
+
+(defn get-bucket-name [uuid]
+      (if-let [[id bucket] (re-matches #"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{8}([0-9a-f]{4})" uuid)]
```
pwojcik86: Use a wildcard `_` instead of `id`, since the value is not used later

PR #146, opened 2017-11-24
SER-1418 | use kubernetes domain instead of consul

idradm:
https://wikia-inc.atlassian.net/browse/SER-1418

Vignette will use kubernetes domain instead of consul discovery.


src/vignette/util/services.clj
```diff
@@ -1,10 +1,13 @@
 (ns vignette.util.services
-  (:require [vignette.util.consul :as consul]
-            [clojure.string :refer [join]]
-            [environ.core :refer [env]]))
+  (:require [clojure.string :refer [join]]
+    [environ.core :refer [env]]))
```
pwojcik86: Please format this line, so that dependencies are aligned

src/vignette/util/services.clj
```diff
@@ -1,10 +1,13 @@
...
+
+
+(defn get-domain []
+  (let [environment (env :wikia-environment)
+        datacenter (env :wikia-datacenter)]
+    (format "http://%s.%s.k8s.wikia.net" environment
```
pwojcik86: I think it will be easier to configure the whole url

src/vignette/util/services.clj
```diff
@@ -1,10 +1,13 @@
...
+      (if (= environment "dev") (join "-" [datacenter environment]) datacenter))))
```
pwojcik86: This can be a value in `let`

src/vignette/util/services.clj
```diff
@@ -1,10 +1,13 @@
 (ns vignette.util.services
-  (:require [vignette.util.consul :as consul]
-            [clojure.string :refer [join]]
-            [environ.core :refer [env]]))
+  (:require [clojure.string :refer [join]]
+    [environ.core :refer [env]]))
+
+
+(defn get-domain []
```
pwojcik86: This function should be inlined in `materialize-static-asset-url`

src/vignette/util/services.clj
```diff
@@ -1,10 +1,12 @@
...
 (defn materialize-static-asset-url [oid]
-  (str
-    (consul/->uri
-      (consul/find-service
-        consul/create-consul "static-assets" consul/service-query-tag)) "/image/" oid))
\ No newline at end of file
+  (let [environment (env :wikia-environment)
+        physical-dc (env :wikia-datacenter)
+        logical-dc (if (= environment "dev") (clojure.string/join "-" [physical-dc environment]) physical-dc)]
+    (if (empty? environment) (throw (Exception. "WIKIA_ENVIRONMENT not set"))
```
pwojcik86: I think it'd be better to move this check to a function and then create constants for `environment` and `logical-dc`, so that the exception is thrown earlier.

PR #152, opened 2018-04-19
SER-1904 | use only the last value from query params

idradm:


src/vignette/util/query_options.clj
```diff
@@ -40,12 +40,14 @@
...
   [request]
-  (reduce (fn [running [key val]]
-            (if (and (contains? query-opts-map (keyword key))
-                     (re-matches (query-opt-regex (get query-opts-map (keyword key))) val))
-              (assoc running (keyword key) val)
-              running))
-          {} (:query-params request)))
+  (reduce
+   (fn [running [key val]]
+     (let [first (first (flatten [val]))]
```
themech: just wondering, why not the `last` value? It is slightly easier to use, as if I want to enforce a format, I can just take any vignette url and append `&format=xyz` at the end without worrying some existing query param takes precedence.

tmszdmsk: @themech, yep, we discussed that it will be the `last` one

PR #168, opened 2018-10-11
SER-2714 | better docker image

tmszdmsk:
modifying docker image so that we use the same versions of libraries as on our "bare metal" instances.

### DIFF
--- a/Dockerfile-k8s
+++ b/Dockerfile-k8s
@@ -22,4 +22,4 @@ EXPOSE 8080
 
 USER service_user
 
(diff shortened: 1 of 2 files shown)


Dockerfile-k8s
```diff
@@ -1,9 +1,18 @@
-FROM alpine:3.8
+FROM ubuntu:14.04.1
 
-RUN adduser -S -D -h /nonexistent -H service_user
+RUN echo "deb [arch=amd64] http://repo.prod.wikia.net/ubuntu trusty main restricted universe multiverse\n" > /etc/apt/sources.list && \
```
harnash: https://help.ubuntu.com/community/add-apt-repository

